PR #16, opened 2026-03-05
Forgot password flow 1st pass

### LINKED ISSUE #13: Feature: Create "Forget Password" flow with Supabase
People who signed up via email and forget their passwords need to be able to retrieve their passwords.

There is a way to do this with Supabase, and set up a link. Need to investigate how this is done, and implement it.

timchaffee:
1st pass at #13 - it works locally but the Vercel dev deploy does not get redirected back to the correct instance from supabase when you click the link in the email. I think that may be a env config issue in Vercel or Supabase. I don't have access to those to check.

### DIFF
--- a/app/(auth)/forgot-password/forgot-password-form.tsx
+++ b/app/(auth)/forgot-password/forgot-password-form.tsx
@@ -1,0 +1,106 @@
+"use client";
+
+import { cn } from "@/lib/utils";
+import { useState } from "react";
+import Link from "next/link";
+import { createClient } from "@/lib/supabase/client";
+import { Button } from "@/components/ui/button";
+import { Input } from "@/components/ui/input";
+import { Label } from "@/components/ui/label";
+import {
+  Card,
+  CardContent,
+  CardDescription,
+  CardHeader,
+  CardTitle,
+} from "@/components/ui/card";
+
+
+export function ForgotPasswordForm({
+  className,
+  ...props
+}: React.ComponentPropsWithoutRef<"div">) {
+  const [email, setEmail] = useState("");
+  const [error, setError] = useState<string | null>(null);
+  const [success, setSuccess] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const handleForgotPassword = async (e: React.FormEvent) => {
+    e.preventDefault();
+    const supabase = createClient();
+    setIsLoading(true);
+    setError(null);
+
+    try {
+      // The url which will be included in the email. This URL needs to be configured in your redirect URLs in the Supabase dashboard at https://supabase.com/dashboard/project/_/auth/url-configuration
+      const { error } = await supabase.auth.resetPasswordForEmail(email, {
+        redirectTo: `${window.location.origin}/update-password`,
+      });
+      if (error) throw error;
+      setSuccess(true);
+    } catch (error: unknown) {
+      setError(error instanceof Error ? error.message : "An error occurred");
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  return (
+    <div className={cn("flex flex-col gap-6", className)} {...props}>
+      {success ? (
+        <Card>
+          <CardHeader>
+            <CardTitle className="text-2xl">Check Your Email</CardTitle>
+            <CardDescription>Password reset instructions sent</CardDescription>
+          </CardHeader>
+          <CardContent>
+            <p className="text-sm text-muted-foreground">
+              If you registered using your email and password, you will receive
+              a password reset email.
+            </p>
+          </CardContent>
+        </Card>
+      ) : (
+        <Card>
+          <CardHeader>
+            <CardTitle className="text-2xl">Reset Your Password</CardTitle>
+            <CardDescription>
+              Type in your email and we&apos;ll send you a link to reset your
+              password
+            </CardDescription>
+          </CardHeader>
+          <CardContent>
+            <form onSubmit={handleForgotPassword}>
+              <div className="flex flex-col gap-6">
+                <div className="grid gap-2">
+                  <Label htmlFor="email">Email</Label>
+                  <Input
+                    id="email"
+                    type="email"
+                    placeholder="m@example.com"
+                    required
+                    value={email}
+                    onChange={(e) => setEmail(e.target.value)}
+                  />
+                </div>
+                {error && <p className="text-sm text-red-500">{error}</p>}
+                <Button type="submit" className="w-full" disabled={isLoading}>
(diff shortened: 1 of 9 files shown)


app/(auth)/forgot-password/forgot-password-form.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+
+  const handleForgotPassword = async (e: React.FormEvent) => {
+    e.preventDefault();
+    const supabase = createClient();
+    setIsLoading(true);
+    setError(null);
+
+    try {
+      // The url which will be included in the email. This URL needs to be configured in your redirect URLs in the Supabase dashboard at https://supabase.com/dashboard/project/_/auth/url-configuration
```
kaizengrowth, line 35: Hi @timchaffee, thanks for noting this. I'll configure this now.

I also added you as an admin on our Supabase account in case you'd like to look at the schema, or make future changes for auth/redirects/config.

kaizengrowth: Here are the two redirects previously configured in Supabase:

Image

Should there be any additional changes here, or an additional URL that is needed? @timchaffee

app/(auth)/update-password/update-password-form.tsx
```diff
@@ -0,0 +1,77 @@
...
+export function UpdatePasswordForm({
+  className,
+  ...props
+}: React.ComponentPropsWithoutRef<"div">) {
+  const [password, setPassword] = useState("");
+  const [error, setError] = useState<string | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const router = useRouter();
+
+  const handleForgotPassword = async (e: React.FormEvent) => {
```
kaizengrowth: NIT: would it be good to rename this to "handleUpdatePassword" for specificity?

app/(auth)/forgot-password/forgot-password-form.tsx
kaizengrowth: Actually realizing that this is insufficient because `forgot-password-form.tsx` sends Supabase this redirect URL:

`${window.location.origin}/update-password`

Our only allowed URLs end in `/auth/callback` so Supabase will reject the redirect, since `update-password` is not in the whitelist. This is why the Vercel preview deploy isn't working.

kaizengrowth: It's also probably not working because the preview deploy for the PR has a different hostname: `community-platform-x74m-git-13-feature-crea-550b3c-what-we-will.vercel.app` rather than: `community-platform-x74m.vercel.app`

I can add a wildcard redirect in our Supabase redirect URL list to allow for preview deploys:
`https://*.vercel.app/auth/callback`

It seems like the cleanest solution (and the standard Supabase + Next.js pattern) is to keep routing through `/auth/callback` and pass the destination as a `next` query param.

app/(auth)/forgot-password/forgot-password-form.tsx
```diff
@@ -0,0 +1,106 @@
...
+    <div className={cn("flex flex-col gap-6", className)} {...props}>
+      {success ? (
+        <Card>
+          <CardHeader>
+            <CardTitle className="text-2xl">Check Your Email</CardTitle>
+            <CardDescription>Password reset instructions sent</CardDescription>
+          </CardHeader>
+          <CardContent>
+            <p className="text-sm text-muted-foreground">
+              If you registered using your email and password, you will receive
```
kaizengrowth: I wonder if a clearer message might be:

"If an account exists for that address, a password reset link will be sent."

app/(auth)/update-password/update-password-form.tsx
```diff
@@ -0,0 +1,77 @@
+"use client";
```
kaizengrowth, line 1: If a user navigates directly to `/update-password` without clicking a reset email link, `supabase.auth.updateUser` will fail with an error. The current error handling will surface that, but it's worth considering redirecting to `/forgot-password` in that case (or checking for a valid session on mount).

app/(auth)/update-password/update-password-form.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <CardDescription>
+            Please enter your new password below.
+          </CardDescription>
+        </CardHeader>
+        <CardContent>
+          <form onSubmit={handleForgotPassword}>
+            <div className="flex flex-col gap-6">
+              <div className="grid gap-2">
+                <Label htmlFor="password">New password</Label>
+                <Input
```
kaizengrowth, line 72: I think this input is missing the attribute:

`autoComplete="new-password"` 

which is in other login forms

app/(auth)/forgot-password/page.tsx
```diff
@@ -0,0 +1,8 @@
+import { ForgotPasswordForm } from "./forgot-password-form";
+
+export default async function ForgotPasswordPage() {
```
kaizengrowth: Does this function need to return async as there's nothing being awaited here?

app/(auth)/forgot-password/forgot-password-form.tsx
timchaffee: Okay making these changes now

app/(auth)/forgot-password/page.tsx
timchaffee: Good catch

app/(auth)/update-password/update-password-form.tsx
timchaffee: How about of there is no CODE QP we redirect to the home page?

app/(auth)/update-password/update-password-form.tsx
timchaffee: Done - naming is very important! 👍

app/(auth)/update-password/update-password-form.tsx
timchaffee: Added

app/(auth)/forgot-password/forgot-password-form.tsx
timchaffee: I have also updated the URLS in Supabase and removed the ones I added for update-password

Image

app/(auth)/update-password/update-password-form.tsx
timchaffee: I made it so if no code QP it goes to forgot password

app/(auth)/update-password/update-password-form.tsx
timchaffee: and added a few more :)

PR #18, opened 2026-03-05
Fix: Online/Offline Status Indicator

### LINKED ISSUE #9: Bug Fix:  Online/Offline indicator does not seem to be working
## 🐛 Bug: Online/Offline Status Indicator Not Working

**Priority:** Medium
**Component:** `UserAvatar`, `last_seen_at` tracking, `getOnlineStatus()`
**Affected areas:** Member cards, conversation list, public profiles, chat headers

---

### Description

The online/offline/away status dots on user avatars do not appear to reflect actual user activity. Users who are actively browsing show as offline, or status does not update when users navigate the app.

### Expected Behavior

- 🟢 **Online** (green dot): `last_seen_at` within the last 5 minutes
- 🟡 **Away** (yellow dot): `last_seen_at` within the last 30 minutes
- ⚫ **Offline** (gray dot): `last_seen_at` older than 30 minutes or null

SirFuzzyWalls:
Note: Both the code and the description of this PR contain AI generated portions, this PR was manually tested and verified by a human, and the description was checked for any obvious mistakes. 

**Problem:**
Two bugs were causing the online/offline status dots to be unreliable:                                            
                                                                                                                    
  1. last_seen_at was never written to the DB. The middleware called supabase.rpc('update_last_seen') but didn't
  await it. In Next.js middleware the response is returned immediately, so the unawaited promise was silently
  dropped — meaning last_seen_at stayed null for every user.
  2. Status dots were stale after the initial render. Even with correct data, getOnlineStatus() was called once at
  server render time and never again. A user shown as online at page load would remain green indefinitely, even
  after 30+ minutes of inactivity.

  **Solution:**

  Part 1 — lib/supabase/proxy.ts
  Added await to the update_last_seen RPC call in middleware so the DB write actually completes before the response
  is returned.

  Part 2 — components/shared/LiveStatusAvatar.tsx (new)
  Created a LiveStatusAvatar client component that wraps UserAvatar and re-evaluates getOnlineStatus() every 60
  seconds using Date.now(). Status dots now decay correctly (online → away → offline) while the user stays on the
  page — no polling or network requests needed, since status is pure time arithmetic on last_seen_at.

  Replaced all getOnlineStatus + UserAvatar usages with LiveStatusAvatar in:
  - /members page cards
  - Group member list
  - DM conversation list sidebar
  - Member profile page (/members/[userId])

  Why 60 seconds?

  The thresholds are 5 min (online → away) and 30 min (away → offline). A 60-second re-evaluation interval is more
  than precise enough and matches the middleware's 1-minute last_seen_at update granularity.

  **Not currently in scope:**

  The reverse direction — a previously-offline user coming back online while you're already on the page — would
  require polling or Supabase Realtime. Lower priority; not currently opened as an issue.

### DIFF
--- a/components/shared/LiveStatusAvatar.tsx
+++ b/components/shared/LiveStatusAvatar.tsx
@@ -22,6 +22,7 @@ export function LiveStatusAvatar({
   );
 
   useEffect(() => {
+    setStatus(getOnlineStatus(lastSeenAt, { isCurrentUser }));
     const id = setInterval(() => {
       setStatus(getOnlineStatus(lastSeenAt, { isCurrentUser }));
     }, 60_000);
--- a/components/shared/LiveStatusAvatar.tsx
+++ b/components/shared/LiveStatusAvatar.tsx
@@ -1,0 +1,32 @@
+"use client";
+
+import { useState, useEffect } from "react";
+import { getOnlineStatus } from "@/lib/utils/status";
+import { UserAvatar } from "./UserAvatar";
+import type { ComponentProps } from "react";
+
+type UserAvatarProps = ComponentProps<typeof UserAvatar>;
+
+interface LiveStatusAvatarProps extends Omit<UserAvatarProps, "status" | "showStatus"> {
+  lastSeenAt: string | null;
+  isCurrentUser?: boolean;
+}
+
+export function LiveStatusAvatar({
+  lastSeenAt,
+  isCurrentUser,
+  ...avatarProps
+}: LiveStatusAvatarProps) {
+  const [status, setStatus] = useState(() =>
+    getOnlineStatus(lastSeenAt, { isCurrentUser })
+  );
+
+  useEffect(() => {
+    const id = setInterval(() => {
+      setStatus(getOnlineStatus(lastSeenAt, { isCurrentUser }));
+    }, 60_000);
+    return () => clearInterval(id);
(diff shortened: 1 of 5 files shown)


lib/supabase/proxy.ts
```diff
@@ -88,7 +88,7 @@ export async function updateSession(request: NextRequest) {
       now - parseInt(lastSeenCookie, 10) > LAST_SEEN_THRESHOLD_MS;
 
     if (shouldUpdate) {
-      void supabase
+      await supabase
```
kaizengrowth, line 110: Nice fix here!

components/shared/LiveStatusAvatar.tsx
```diff
@@ -0,0 +1,32 @@
...
+  lastSeenAt,
+  isCurrentUser,
+  ...avatarProps
+}: LiveStatusAvatarProps) {
+  const [status, setStatus] = useState(() =>
+    getOnlineStatus(lastSeenAt, { isCurrentUser })
+  );
+
+  useEffect(() => {
+    const id = setInterval(() => {
```
kaizengrowth, line 26: it would be good to add here below line 24, before the `setInterval`

`setStatus(getOnlineStatus(lastSeenAt, { isCurrentUser }));`

kaizengrowth (approved):

PR #22, opened 2026-03-07
first pass at claude and cursor boundaries

SirFuzzyWalls:
Adds baseline ignore configs for Cursor and Claude Code to keep sensitive files out of AI context.

**What's included:**
- `.cursorignore` — stops Cursor from indexing secrets, Supabase local config, keys, and logs - _best effort according to cursor_
- `.claude/settings.json` — deny rules for Claude Code covering the same paths

Snippet from Claude CLI when asked to access .env.local (do we believe Claude is incapable of gaslighting?)

" Your permission settings are blocking me from reading .env.local. This is likely intentional — .env.local         
  typically contains sensitive secrets like API keys, database credentials, and tokens that you may not want
  exposed.                                                                                                          
                                                                                                                    
  If you'd like me to read it, you can adjust your permission settings to allow access to that file. "

**Note**: Both files are intentionally lean for now, plans to expand around group calls and user verification features in the future.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,13 @@
+{
+  "permissions": {
+    "deny": [
+      "Read(./.env)",
+      "Read(./.env.*)",
+      "Read(./.env.local)",
+      "Read(./supabase/.temp/**)",
+      "Read(./.supabase/**)",
+      "Read(./.next/**)",
+      "Read(./node_modules/**)"
+    ]
+  }
+}
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,13 +1,13 @@
 {
   "permissions": {
     "deny": [
-      "Read(./.env)",
+      "Read(./.env*)",
       "Read(./.env.*)",
-      "Read(./.env.local)",
       "Read(./supabase/.temp/**)",
       "Read(./.supabase/**)",
       "Read(./.next/**)",
-      "Read(./node_modules/**)"
+      "Read(./node_modules/**)",
+      "Read(./tsconfig.tsbuildinfo)" 
     ]
   }
 }
--- a/.cursorignore
+++ b/.cursorignore
@@ -1,5 +1,4 @@
 # Secrets & env
-.env.local
 .env
 .env.*
 
--- a/.cursorignore
+++ b/.cursorignore
@@ -1,0 +1,15 @@
+# Secrets & env
+.env.local
+.env
+.env.*
+
+# Supabase
+supabase/.temp/
+.supabase/
+
+# Build & deps
+.next/
+node_modules/
+
+# TypeScript incremental build info (large, not useful for AI)
+tsconfig.tsbuildinfo
--- a/AI_POLICY.md
+++ b/AI_POLICY.md
@@ -1,0 +1,68 @@
+# AI Use Policy
+
+Welcome! We are open to the use of AI tools so everyone can contribute effectively regardless of experience level. This policy exists in an effort to reduce technical risks for our contributors, and hopefully refuce downtime for our members, not to overly restrict or police the tools used.
+
+**Mistakes happen:** If you make a mistake involving an AI tool such as: accidentally sharing sensitive data, pasting the wrong thing, or leaking PII,  please report it immediately. Mistakes happen, reporting them makes us all more aware & mindful in the future. After a report, appropriate steps will be taken to rotate keys, alert affected individuals, and enhance AI use guidelines.
+
+**How to report a breach:** The best way to is to DM a maintainer or mod via Slack. If you're not a member of our Slack, please open Github issue titled "AI breach" : "context" and request a mod or maintainer review.
+ 
+---
+
+## Pull request policy
+Disclose the AI tools used, and to what degree. 
+
+---
+
+## Approved tools
+
+- **Cursor** (with pre-configured `.cursorignore`)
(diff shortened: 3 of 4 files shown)


kaizengrowth (commented):
Hey Simon! Thanks for these Cursor setting suggestions. I think these are all really solid. 

Making sure Cursor does not have access to our env variables and access keys seems like a necessary place to start.

SirFuzzyWalls:
@kaizengrowth @timchaffee @kylealbasi Would y'all mind taking a look at `AI_POLICY.md`? It's very early and I'd like to start the discussion more formally for all of us. Thank you!

.claude/settings.json
```diff
@@ -0,0 +1,13 @@
+{
+  "permissions": {
+    "deny": [
+      "Read(./.env)",
+      "Read(./.env.*)",
+      "Read(./.env.local)",
```
timchaffee: These seem duplicative? Won't 

` "Read(./.env.*)",`

block 

`"Read(./.env.local)"`

.claude/settings.json
```diff
@@ -0,0 +1,13 @@
...
+      "Read(./supabase/.temp/**)",
+      "Read(./.supabase/**)",
+      "Read(./.next/**)",
+      "Read(./node_modules/**)"
```
timchaffee: I am curious as to why we would now want claude to see in node modules? Not suggesting we allow it - it can see what packages we have via packages.json so it should not need this.

.cursorignore
```diff
@@ -0,0 +1,15 @@
+# Secrets & env
+.env.local
+.env
+.env.*
```
timchaffee, line 3: Again doesn't `.env.*` also block `.env.local`?

.cursorignore
```diff
@@ -0,0 +1,15 @@
...
+# Supabase
+supabase/.temp/
+.supabase/
+
+# Build & deps
+.next/
+node_modules/
+
+# TypeScript incremental build info (large, not useful for AI)
+tsconfig.tsbuildinfo
```
timchaffee, line 14: Should this be added to the Claude ignore?

AI_POLICY.md
```diff
@@ -0,0 +1,68 @@
+# AI Use Policy
+
+Welcome! We are open to the use of AI tools so everyone can contribute effectively regardless of experience level. This policy exists in an effort to reduce technical risks for our contributors, and hopefully refuce downtime for our members, not to overly restrict or police the tools used.
+
+**Mistakes happen:** If you make a mistake involving an AI tool such as: accidentally sharing sensitive data, pasting the wrong thing, or leaking PII,  please report it immediately. Mistakes happen, reporting them makes us all more aware & mindful in the future. After a report, appropriate steps will be taken to rotate keys, alert affected individuals, and enhance AI use guidelines.
```
timchaffee: Nice - maybe add "We all make mistakes, acknowledging them and learning from them is what we care about!

AI_POLICY.md
```diff
@@ -0,0 +1,68 @@
...
+- **Cursor** (with pre-configured `.cursorignore`)
+- **Claude Code** (with pre-configured `.claude/settings.json`)
+
+We can't cover every AI tool that emerges, so please request approval for use before using a tool not listed above.Different tools have different data retention and privacy policies, and we want to understand the policies of each tool in use. Please open an issue on Github or reach out via slack to a mod or maintainer for tool evaluation and sign off!
+
+---
+
+## Tasks that AI tools are generally helpful for:
+- Generating boilerplate, scaffolding, and repetitive code
+- Writing and improving tests
```
timchaffee, line 27: This is a great place to start.

AI_POLICY.md
```diff
@@ -0,0 +1,68 @@
...
+---
+
+## Keeping secrets safe
+
+- Never paste `.env` contents, API keys, or credentials into an AI chat.
+- The `.cursorignore` and `.claude/settings.json` configs in this repo are there to help you — please don't remove or weaken them.
+
+We will use separate Supabase projects for dev/staging/production. As a volunteer, you should only ever need the dev and staging credentials. If someone asks you to use production credentials, please check in with a maintainer or mod before proceeding. 
+
+---
```
timchaffee, line 62: I think we should add a section or statement to the effect: **NEVER give AI direct access to the GitHub repository. No AI agents are allowed in GitHub or in automating anything related to GitHub.**

timchaffee (approved):
Looks really good. I have a few comments but its all minor.

.claude/settings.json
kaizengrowth: Yes, doesn't seem necessary to block reading node_modules

AI_POLICY.md
kaizengrowth: I like that! :)

AI_POLICY.md
kaizengrowth: I've definitely been guilty of using Claude Code for more than boilerplate in a rush. However, this is really good guidance here.

AI_POLICY.md
kaizengrowth: Yes

.claude/settings.json
SirFuzzyWalls: You're correct! These are boilerplate examples, afaik there isn't any issue here, but it's not clean.

I'll leave the first two lines and remove line 6 here and in the other file with a similar structure.

PR #28, opened 2026-03-14
Fix approvals page not refreshing and pending members showing on dash…

SirFuzzyWalls:
**AI Use:** Claude code for local review & some code generation. 

Additional changes: 
  - `actions.ts`: Extracted duplicated admin auth check into ensureAdmin() helper, shared by both functions.
  - `actions.ts`: Parallelized the profile update and email lookup in approveUser with Promise.all — they're
  independent operations.
  - `NewMembersCard.tsx`: Replaced select("*") with the 5 fields actually used (id, created_at, avatar_url,
  display_name, headline).
  - `NewMembersCard.tsx`: Moved the 30-day filter into the query as .gte("created_at", ...) so the limit(10) applies
  to the correctly filtered set. Also fixed the stale module-level THIRTY_DAYS_AGO constant (it was computed once at
   startup, not per request).

### DIFF
--- a/app/admin/approvals/actions.ts
+++ b/app/admin/approvals/actions.ts
@@ -1,21 +1,20 @@
 "use server";
 
 import { redirect } from "next/navigation";
+import { revalidatePath } from "next/cache";
 import { createClient } from "@/lib/supabase/server";
 import { createServiceClient } from "@/lib/supabase/service";
 import nodemailer from "nodemailer";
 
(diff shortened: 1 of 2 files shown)


kaizengrowth:
Excellent! Thanks for making this fix! @SirFuzzyWalls

kaizengrowth (approved):

Issue #3, opened 2026-03-02
Feature: Group shared notes tab


kaizengrowth:
**Context:** The Group hub (e.g. `/groups/[slug]`) has a "Notes" tab that currently shows a placeholder: "Shared notes coming soon".

**Goal:** Implement shared notes for groups so members can collaborate on notes (e.g. meeting notes, resources). Consider:
- Data model (notes per group, versioning?)
- Notes tab content: list/create/edit/delete notes
- Permissions (who can edit: all members vs admins only)

**References:** `app/(app)/groups/[slug]/GroupHubClient.tsx` (Notes TabsContent, ~line 376).

PR #36, opened 2026-03-20
Fixes #31 bug time zone setting in events listings urgent

### LINKED ISSUE #31: Bug: Time zone setting in events listings (urgent)
Asana Link: https://app.asana.com/1/1213515485882011/project/1213516055535837/task/1213746764285240?focus=true
Some really weird things happening with time zone settings in our Events calendar:

Image

translates to:

Image

tonyrosario:
pushing up the hackathon from Friday.  Primary goal was to get an idea on the scope of changes, not ready to be submitted yet.

I did have some trouble with the migrations, which is why they are modified.  They don't need to be a part of this PR.

### DIFF
--- a/app/(app)/app-shell.tsx
+++ b/app/(app)/app-shell.tsx
@@ -35,6 +35,7 @@ import { cn } from "@/lib/utils";
 import { UnreadBadge } from "@/components/messages/UnreadBadge";
 import { UserAvatar } from "@/components/shared/UserAvatar";
 import { updateLastSeen } from "@/app/(app)/profile/actions";
+import { syncBrowserTimezone } from "@/lib/actions/timezone";
 
 const HEARTBEAT_INTERVAL_MS = 45_000; // 45s — keep last_seen_at fresh so others see you online
 
@@ -85,6 +86,17 @@ export default function AppShell({ children, user }: AppShellProps) {
     }, HEARTBEAT_INTERVAL_MS);
     updateLastSeen(); // run once on mount
     return () => clearInterval(interval);
+  }, []);
+
+  // One-time timezone sync: if the user's profile still has the default
+  // timezone, update it to match the browser's detected timezone.
+  useEffect(() => {
+    if (sessionStorage.getItem("tz_checked")) return;
+    sessionStorage.setItem("tz_checked", "1");
+    const browserTz = Intl.DateTimeFormat().resolvedOptions().timeZone;
+    if (browserTz) {
+      syncBrowserTimezone(browserTz);
+    }
   }, []);
 
   async function handleSignOut() {
--- a/app/(app)/events/[eventId]/edit/page.tsx
+++ b/app/(app)/events/[eventId]/edit/page.tsx
@@ -21,7 +21,7 @@ export default async function EditEventPage({
   const { data: event } = await supabase
     .from("events")
     .select(
-      "id, title, description, event_type, host_id, starts_at, ends_at, location, max_attendees, group_id"
+      "id, title, description, event_type, host_id, starts_at, ends_at, location, max_attendees, group_id, timezone"
     )
     .eq("id", eventId)
     .single();
@@ -63,6 +63,7 @@ export default async function EditEventPage({
           location: event.location,
           max_attendees: event.max_attendees,
           group_id: event.group_id,
+          timezone: event.timezone ?? "America/Chicago",
         }}
         groups={groups}
       />
--- a/app/(app)/events/[eventId]/page.tsx
+++ b/app/(app)/events/[eventId]/page.tsx
@@ -1,7 +1,7 @@
 import { notFound, redirect } from "next/navigation";
 import Link from "next/link";
 import { createClient } from "@/lib/supabase/server";
-import { formatInTimeZone, getTimeZoneAbbreviation } from "@/lib/utils/timezone";
+import { EventTimeDisplay } from "@/components/events/EventTimeDisplay";
 import {
   ArrowLeft,
   Video,
@@ -65,8 +65,6 @@ export default async function EventDetailPage({
   };
 
   const eventTz = event.timezone ?? viewerTimezone;
-  const sameTz = eventTz === viewerTimezone;
-  const eventTzAbbr = getTimeZoneAbbreviation(event.starts_at, eventTz);
 
   const { data: attendees } = await supabase
     .from("event_rsvps")
@@ -114,20 +112,15 @@ export default async function EventDetailPage({
             {event.title}
           </h1>
           <div className="mt-2 flex flex-wrap items-center gap-x-4 gap-y-1 text-muted-foreground">
-            <div className="flex flex-col gap-0.5">
-              <span className="flex items-center gap-1.5">
-                <Calendar className="size-4 shrink-0" />
-                {formatInTimeZone(event.starts_at, eventTz, "EEEE, MMMM d, yyyy")} ·{" "}
-                {formatInTimeZone(event.starts_at, eventTz, "h:mm a")} –{" "}
-                {formatInTimeZone(event.ends_at, eventTz, "h:mm a")} {eventTzAbbr}
-              </span>
-              {!sameTz && (
-                <span className="ml-5.5 text-sm text-muted-foreground/70">
-                  {formatInTimeZone(event.starts_at, viewerTimezone, "h:mm a")} –{" "}
-                  {formatInTimeZone(event.ends_at, viewerTimezone, "h:mm a")}{" "}
-                  {getTimeZoneAbbreviation(event.starts_at, viewerTimezone)} in your time
-                </span>
-              )}
+            <div className="flex items-start gap-1.5">
+              <Calendar className="size-4 shrink-0 mt-0.5" />
+              <EventTimeDisplay
+                startsAt={event.starts_at}
+                endsAt={event.ends_at}
+                eventTimezone={eventTz}
+                profileTimezone={viewerTimezone}
+                dateFormat="EEEE, MMMM d, yyyy"
+              />
             </div>
             {event.location && (
               <span className="flex items-center gap-1.5">
--- a/app/(app)/events/[eventId]/page.tsx
+++ b/app/(app)/events/[eventId]/page.tsx
@@ -1,13 +1,12 @@
 import { notFound, redirect } from "next/navigation";
 import Link from "next/link";
-import { format } from "date-fns";
 import { createClient } from "@/lib/supabase/server";
+import { formatInTimeZone } from "@/lib/utils/timezone";
 import {
   ArrowLeft,
   Video,
   Calendar,
   MapPin,
-  User,
 } from "lucide-react";
 import { Button } from "@/components/ui/button";
 import { Card, CardContent } from "@/components/ui/card";
@@ -33,6 +32,13 @@ export default async function EventDetailPage({
     data: { user },
   } = await supabase.auth.getUser();
   if (!user) redirect("/login");
+
+  const { data: viewerProfile } = await supabase
+    .from("profiles")
+    .select("timezone")
+    .eq("id", user.id)
+    .single();
+  const viewerTimezone = viewerProfile?.timezone ?? "America/Chicago";
 
   const eventData = await fetchEventWithDetails(eventId, user.id);
   if (!eventData) notFound();
@@ -65,7 +71,6 @@ export default async function EventDetailPage({
     .order("created_at", { ascending: true });
 
   const startsAt = new Date(event.starts_at);
-  const endsAt = new Date(event.ends_at);
   const now = new Date();
   const minsToStart = Math.round((startsAt.getTime() - now.getTime()) / 60000);
   const startingSoon =
@@ -106,7 +111,9 @@ export default async function EventDetailPage({
           <div className="mt-2 flex flex-wrap items-center gap-x-4 gap-y-1 text-muted-foreground">
             <span className="flex items-center gap-1.5">
               <Calendar className="size-4 shrink-0" />
-              {format(startsAt, "EEEE, MMMM d, yyyy")} · {format(startsAt, "h:mm a")} – {format(endsAt, "h:mm a")}
+              {formatInTimeZone(event.starts_at, viewerTimezone, "EEEE, MMMM d, yyyy")} ·{" "}
+              {formatInTimeZone(event.starts_at, viewerTimezone, "h:mm a")} –{" "}
+              {formatInTimeZone(event.ends_at, viewerTimezone, "h:mm a")}
             </span>
             {event.location && (
               <span className="flex items-center gap-1.5">
--- a/app/(app)/events/[eventId]/page.tsx
+++ b/app/(app)/events/[eventId]/page.tsx
@@ -1,7 +1,7 @@
 import { notFound, redirect } from "next/navigation";
 import Link from "next/link";
 import { createClient } from "@/lib/supabase/server";
-import { formatInTimeZone } from "@/lib/utils/timezone";
+import { formatInTimeZone, getTimeZoneAbbreviation } from "@/lib/utils/timezone";
 import {
   ArrowLeft,
   Video,
@@ -55,6 +55,7 @@ export default async function EventDetailPage({
     starts_at: string;
     ends_at: string;
     max_attendees: number | null;
+    timezone: string;
     host: Profile | null;
     group: unknown;
     rsvpCounts: { going: number; maybe: number; declined: number };
@@ -62,6 +63,10 @@ export default async function EventDetailPage({
     isLive: boolean;
     isPast: boolean;
   };
+
+  const eventTz = event.timezone ?? viewerTimezone;
+  const sameTz = eventTz === viewerTimezone;
+  const eventTzAbbr = getTimeZoneAbbreviation(event.starts_at, eventTz);
 
   const { data: attendees } = await supabase
     .from("event_rsvps")
@@ -109,12 +114,21 @@ export default async function EventDetailPage({
             {event.title}
           </h1>
           <div className="mt-2 flex flex-wrap items-center gap-x-4 gap-y-1 text-muted-foreground">
-            <span className="flex items-center gap-1.5">
-              <Calendar className="size-4 shrink-0" />
-              {formatInTimeZone(event.starts_at, viewerTimezone, "EEEE, MMMM d, yyyy")} ·{" "}
-              {formatInTimeZone(event.starts_at, viewerTimezone, "h:mm a")} –{" "}
-              {formatInTimeZone(event.ends_at, viewerTimezone, "h:mm a")}
-            </span>
+            <div className="flex flex-col gap-0.5">
+              <span className="flex items-center gap-1.5">
+                <Calendar className="size-4 shrink-0" />
+                {formatInTimeZone(event.starts_at, eventTz, "EEEE, MMMM d, yyyy")} ·{" "}
+                {formatInTimeZone(event.starts_at, eventTz, "h:mm a")} –{" "}
+                {formatInTimeZone(event.ends_at, eventTz, "h:mm a")} {eventTzAbbr}
+              </span>
+              {!sameTz && (
+                <span className="ml-5.5 text-sm text-muted-foreground/70">
+                  {formatInTimeZone(event.starts_at, viewerTimezone, "h:mm a")} –{" "}
+                  {formatInTimeZone(event.ends_at, viewerTimezone, "h:mm a")}{" "}
+                  {getTimeZoneAbbreviation(event.starts_at, viewerTimezone)} in your time
+                </span>
+              )}
+            </div>
             {event.location && (
               <span className="flex items-center gap-1.5">
                 <MapPin className="size-4 shrink-0" />
--- a/app/(app)/events/actions.ts
+++ b/app/(app)/events/actions.ts
@@ -70,6 +70,7 @@ export async function createEventAction(formData: {
   location: string;
   max_attendees: number | null;
   group_id: string | null;
+  timezone: string;
   recurrence_rule?: "daily" | "weekly" | null;
   recurrence_end_date?: string | null;
 }) {
@@ -81,10 +82,13 @@ export async function createEventAction(formData: {
 
   const { recurrence_rule, recurrence_end_date, ...baseData } = formData;
 
+  const eventTimezone = formData.timezone;
+
   // Create the parent (first) event
   const event = await createEvent({
     ...baseData,
     host_id: user.id,
+    timezone: eventTimezone,
     recurrence_rule: recurrence_rule ?? null,
     recurrence_end_date: recurrence_end_date ?? null,
   });
@@ -96,6 +100,7 @@ export async function createEventAction(formData: {
       formData.ends_at,
       recurrence_rule,
       recurrence_end_date,
+      eventTimezone,
     );
 
     const instances = datePairs.map(({ starts_at, ends_at }) => {
@@ -112,6 +117,7 @@ export async function createEventAction(formData: {
         starts_at,
         ends_at,
         video_room_name: getVideoRoomName({ type: "event", id }),
+        timezone: eventTimezone,
         recurrence_rule,
         parent_event_id: event.id,
       };
@@ -147,6 +153,7 @@ export async function updateEventAction(
     location: string;
     max_attendees: number | null;
     group_id: string | null;
+    timezone?: string;
   }
 ) {
   const supabase = await createClient();
@@ -176,6 +183,7 @@ export async function updateEventAction(
       location: formData.location.trim() || "Online",
       max_attendees: formData.max_attendees,
       group_id: formData.group_id,
+      ...(formData.timezone ? { timezone: formData.timezone } : {}),
       updated_at: new Date().toISOString(),
     })
     .eq("id", eventId)
--- a/app/(app)/events/create/page.tsx
+++ b/app/(app)/events/create/page.tsx
@@ -16,6 +16,12 @@ export default async function CreateEventPage({
   } = await supabase.auth.getUser();
   if (!user) redirect("/login");
 
+  const { data: profile } = await supabase
+    .from("profiles")
+    .select("timezone")
+    .eq("id", user.id)
+    .single();
+
   const params = await searchParams;
   const preselectedGroupId = params.group ?? null;
 
@@ -46,6 +52,7 @@ export default async function CreateEventPage({
       <CreateEventForm
         groups={groups}
         preselectedGroupId={preselectedGroupId}
+        profileTimezone={profile?.timezone ?? "America/Chicago"}
       />
     </div>
   );
--- a/app/(app)/events/page.tsx
+++ b/app/(app)/events/page.tsx
@@ -2,7 +2,6 @@ import { createClient } from "@/lib/supabase/server";
 import { redirect } from "next/navigation";
 import Link from "next/link";
 import {
-  format,
   isToday,
   isSameDay,
   addDays,
@@ -86,6 +85,13 @@ export default async function EventsPage({
   } = await supabase.auth.getUser();
   if (!user) redirect("/login");
 
+  const { data: viewerProfile } = await supabase
+    .from("profiles")
+    .select("timezone")
+    .eq("id", user.id)
+    .single();
+  const viewerTimezone = viewerProfile?.timezone ?? "America/Chicago";
+
   const [upcomingRaw, pastRaw] = await Promise.all([
     fetchUpcomingEvents({ groupId: null, type: typeParam }),
     fetchPastEvents({ limit: 10 }),
@@ -162,6 +168,7 @@ export default async function EventsPage({
         <EventsCalendarClient
           events={upcomingWithDetails}
           currentUserId={user.id}
+          viewerTimezone={viewerTimezone}
         />
       ) : (
         <>
@@ -179,6 +186,7 @@ export default async function EventsPage({
                         rsvpCounts={event.rsvpCounts}
                         currentUserRsvp={event.currentUserRsvp}
                         currentUserId={user.id}
+                        viewerTimezone={viewerTimezone}
                       />
                     </li>
                   ))}
@@ -187,7 +195,7 @@ export default async function EventsPage({
             ))}
           </div>
 
-          <PastEventsSection past={past} currentUserId={user.id} />
+          <PastEventsSection past={past} currentUserId={user.id} viewerTimezone={viewerTimezone} />
         </>
       )}
     </div>
@@ -197,9 +205,11 @@ export default async function EventsPage({
 function PastEventsSection({
   past,
   currentUserId,
+  viewerTimezone,
 }: {
   past: EventRow[];
   currentUserId: string;
+  viewerTimezone: string;
 }) {
   return (
     <details className="group">
@@ -214,6 +224,7 @@ function PastEventsSection({
               rsvpCounts={{ going: 0, maybe: 0, declined: 0 }}
               currentUserRsvp={null}
               currentUserId={currentUserId}
+              viewerTimezone={viewerTimezone}
             />
           </li>
         ))}
--- a/app/(app)/events/page.tsx
+++ b/app/(app)/events/page.tsx
@@ -33,6 +33,7 @@ type EventRow = {
   max_attendees: number | null;
   created_at: string;
   updated_at: string;
+  timezone?: string;
   host: Profile | null;
 };
 
--- a/app/(app)/groups/[slug]/GroupHubClient.tsx
+++ b/app/(app)/groups/[slug]/GroupHubClient.tsx
@@ -434,6 +434,7 @@ export function GroupHubClient({
                   max_attendees: number | null;
                   created_at: string;
                   updated_at: string;
+                  timezone?: string;
                   host: Profile | null;
                 };
                 return (
--- a/app/(app)/groups/[slug]/GroupHubClient.tsx
+++ b/app/(app)/groups/[slug]/GroupHubClient.tsx
@@ -51,6 +51,7 @@ interface GroupHubClientProps {
   eventRsvpCounts: Record<string, { going: number; maybe: number; declined: number }>;
   eventUserRsvp: Record<string, { status: string }>;
   initialNotes: GroupNote[];
+  viewerTimezone: string;
 }
 
 export function GroupHubClient({
@@ -66,6 +67,7 @@ export function GroupHubClient({
   eventRsvpCounts,
   eventUserRsvp,
   initialNotes,
+  viewerTimezone,
 }: GroupHubClientProps) {
   const router = useRouter();
   const [joining, setJoining] = useState(false);
@@ -441,6 +443,7 @@ export function GroupHubClient({
                       rsvpCounts={eventRsvpCounts[e.id] ?? { going: 0, maybe: 0, declined: 0 }}
                       currentUserRsvp={eventUserRsvp[e.id] ? { event_id: e.id, user_id: currentUser.id, status: eventUserRsvp[e.id].status as "going" | "maybe" | "declined", created_at: "" } : null}
                       currentUserId={currentUser.id}
+                      viewerTimezone={viewerTimezone}
                     />
                   </li>
                 );
--- a/app/(app)/groups/[slug]/page.tsx
+++ b/app/(app)/groups/[slug]/page.tsx
@@ -103,7 +103,7 @@ export default async function GroupHubPage({ params }: Props) {
   // Current user's profile (include role for platform admin features)
   const { data: currentUserProfile } = await supabase
     .from("profiles")
-    .select("id, display_name, avatar_url, role")
+    .select("id, display_name, avatar_url, role, timezone")
     .eq("id", user.id)
     .single();
 
@@ -200,6 +200,7 @@ export default async function GroupHubPage({ params }: Props) {
       eventRsvpCounts={eventRsvpCounts}
       eventUserRsvp={eventUserRsvp}
       initialNotes={notes}
+      viewerTimezone={currentUserProfile?.timezone ?? "America/Chicago"}
     />
   );
 }
--- a/app/(app)/profile/actions.ts
+++ b/app/(app)/profile/actions.ts
@@ -16,6 +16,7 @@ export async function updateProfile(
     linkedin_url?: string | null;
     github_url?: string | null;
     portfolio_url?: string | null;
+    timezone?: string | null;
   }
 ): Promise<ProfileUpdateResult> {
   const supabase = await createClient();
@@ -41,6 +42,7 @@ export async function updateProfile(
       linkedin_url: data.linkedin_url || null,
       github_url: data.github_url || null,
       portfolio_url: data.portfolio_url || null,
+      timezone: data.timezone || "America/Chicago",
       is_onboarded: true,
     },
     { onConflict: "id" }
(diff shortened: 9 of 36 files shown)


tonyrosario:
Here is the plan I generated with Claude:
# Plan: Fix Event Timezone Handling

**Branch:** `fix/event-timezone-handling`
**Prerequisite:** `fix/profile-timezone-detection` (see `plans/local/prereq-timezone-detection.md`) — must ship first so `profiles.timezone` values are accurate before this PR makes them load-bearing.
**Problem:** Events are created and displayed using the browser's local timezone. The user's profile timezone is ignored. Times shift when users travel or have mismatched browser/profile timezones.
**Decision:** Creator's profile timezone is the source of truth. Store it on the event row for display context.

---

## Design

### How it should work

1. **Creation:** User enters "2:00 PM" → system reads their `profile.timezone` (e.g. `America/New_York`) → converts "2:00 PM ET" to UTC → stores UTC in `starts_at`/`ends_at` + stores `"America/New_York"` in `events.timezone`
2. **Display:** Read UTC from DB → convert to the **viewer's profile timezone** → render with the event's original TZ label for context (e.g. "2:00 PM ET · 11:00 AM your time")
3. **Editing:** Load UTC from DB → convert to the **event's stored timezone** for the form fields → on save, convert back to UTC using the event's timezone

### Key constraint
`TIMESTAMPTZ` columns stay as-is. UTC storage is correct. The fix is entirely in how we **convert to/from UTC** on the client and server boundaries.

---

## Steps

### Step 1 — Add `date-fns-tz` dependency

**File:** `package.json`

```bash
npm install date-fns-tz
```

`date-fns-tz` provides `fromZonedTime` (formerly `zonedTimeToUtc`) and `toZonedTime` (formerly `utcToZonedTime`) plus `formatInTimeZone`. This is the standard companion to `date-fns` v3 for IANA timezone work.

---

### Step 2 — Migration: add `timezone` column to `events`

**New file:** `supabase/migrations/057_event_timezone.sql`

```sql
-- Store the IANA timezone the event was created in.
-- Used for display context ("2 PM ET") and correct round-tripping in edit forms.
-- Defaults to America/Chicago to match the profile timezone default.

ALTER TABLE public.events
  ADD COLUMN IF NOT EXISTS timezone TEXT NOT NULL DEFAULT 'America/Chicago';

-- Backfill: set existing events to their host's profile timezone (best guess).
UPDATE public.events e
SET timezone = COALESCE(
  (SELECT p.timezone FROM public.profiles p WHERE p.id = e.host_id),
  'America/Chicago'
);
```

**Why default to America/Chicago:** Matches the `profiles.timezone` default. Existing events were likely created by users whose browser timezone matched this default. Not perfect, but the best backfill available.

---

### Step 3 — Create a shared timezone utility

**New file:** `lib/utils/timezone.ts`

Provides three functions used by all event code.

**Parameter order convention:** All functions use **date-first, timezone-second** — matching the `date-fns-tz` library convention (`formatInTimeZone(date, timeZone, formatStr)`). This avoids confusion for anyone familiar with the underlying library.

```typescript
import {
  fromZonedTime,
  formatInTimeZone as dateFnsFormatInTimeZone,
} from "date-fns-tz";

const DEFAULT_TIMEZONE = "America/Chicago";

/**
 * Validate an IANA timezone string. Returns the timezone if valid,
 * or the default timezone as a safe fallback.
 */
function safeTimezone(tz: string | null | undefined): string {
  if (!tz) return DEFAULT_TIMEZONE;
  try {
    Intl.DateTimeFormat(undefined, { timeZone: tz });
    return tz;
  } catch {
    return DEFAULT_TIMEZONE;
  }
}

/**
 * Convert a naive date + time string (from an input field) in a given
 * IANA timezone to a UTC ISO string for storage.
 *
 * Example: localTimeToUTC("2026-03-26", "14:00", "America/New_York")
 *        → "2026-03-26T18:00:00.000Z"
 */
export function localTimeToUTC(date: string, time: string, tz: string): string {
  return fromZonedTime(`${date}T${time}`, safeTimezone(tz)).toISOString();
}

/**
 * Format a UTC ISO string in a specific IANA timezone.
 * Parameter order matches date-fns-tz: (date, timezone, format).
 *
 * Example: formatInTimeZone("2026-03-26T18:00:00Z", "America/New_York", "h:mm a")
 *        → "2:00 PM"
 */
export function formatInTimeZone(
  utcISO: string,
  tz: string,
  fmt: string
): string {
  return dateFnsFormatInTimeZone(utcISO, safeTimezone(tz), fmt);
}

/**
 * Get the short timezone abbreviation for display.
 * Parameter order: date-first, timezone-second (consistent with formatInTimeZone).
 *
 * Example: getTimeZoneAbbreviation("2026-03-26T18:00:00Z", "America/New_York") → "EDT"
 */
export function getTimeZoneAbbreviation(utcISO: string, tz: string): string {
  return dateFnsFormatInTimeZone(utcISO, safeTimezone(tz), "zzz");
}
```

**Design decisions:**
- **Parameter order** matches `date-fns-tz` (`date, tz, fmt`) to avoid footgun for anyone familiar with the library.
- **`safeTimezone` guard** validates IANA strings and falls back to `America/Chicago` for null, empty, or invalid values. One corrupted profile row won't crash every event page.
- **Centralized wrapper** so components never import `date-fns-tz` directly. If we swap libraries, one file changes.

---

### Step 4 — Update `CreateEventForm.tsx`

**File:** `components/events/CreateEventForm.tsx`

Changes:
1. Add `profileTimezone` as a **required** prop (not optional, no fallback) so `npm run build` fails if any parent omits it
2. Replace `new Date(\`${date}T${startTime}\`).toISOString()` with `localTimeToUTC(date, startTime, profileTimezone)`
3. Same for `endTime`
4. Pass the profile timezone as `timezone` in the form data sent to the server action
5. Update the "Times are in {timezone}" label to show the profile timezone, not the browser timezone

**Diff summary:**
```
  interface CreateEventFormProps {
+   profileTimezone: string;  // REQUIRED — no fallback, build fails if missing
    // ...existing props
  }

- const timezone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+ // profileTimezone comes from props (server component passes it)

- const starts_at = start.toISOString();
- const ends_at = end.toISOString();
+ const starts_at = localTimeToUTC(date, startTime, profileTimezone);
+ const ends_at = localTimeToUTC(date, endTime, profileTimezone);

  // Include timezone in form data
+ timezone: profileTimezone,
```

**Parent server components:** Grep for every render of `<CreateEventForm` in the codebase. Each parent must pass `profileTimezone={profile.timezone}`. Most already fetch the user's profile for auth — the timezone is available on the same object. Since the prop is required, any missed parent will cause a TypeScript build error, making this impossible to ship broken.

**Implementation checklist for this step:**
```bash
# Run this to find every render site:
grep -rn "CreateEventForm" app/ components/
# Every match must pass profileTimezone. No exceptions.
```

---

### Step 5 — Update `EditEventForm.tsx`

**File:** `components/events/EditEventForm.tsx`

Changes:
1. Use the **event's stored timezone** (not browser timezone) to populate form fields
2. Replace `format(startsAt, "yyyy-MM-dd")` / `format(startsAt, "HH:mm")` with `formatInTimeZone(event.starts_at, event.timezone, "yyyy-MM-dd")` and `formatInTimeZone(event.starts_at, event.timezone, "HH:mm")`
3. On save, convert back to UTC using the event's timezone: `localTimeToUTC(date, startTime, event.timezone)`
4. Display "Times are in {event.timezone}" so the editor knows the reference timezone

**Why use event timezone, not profile:** The editor should see times in the timezone the event was originally created in. If someone in PST edits an ET event, the form should show ET times.

---

### Step 6 — Update `createEvent` in `lib/events.ts`

**File:** `lib/events.ts`

Changes:
- Accept `timezone` in the `data` parameter
- Pass it through to the Supabase insert

```diff
  const { data: event, error } = await supabase
    .from("events")
    .insert({
      id: eventId,
      ...data,
+     timezone: data.timezone,
      // ... rest unchanged
    })
```

The UTC conversion now happens in the form (Step 4), so `starts_at` and `ends_at` arrive already as correct UTC strings. The server helper just passes them through.

---

### Step 7 — Update event display components

**Files to change (all known display surfaces):**
- `components/events/EventCard.tsx` — primary event card used in lists
- `components/events/EventsCalendarClient.tsx` — calendar view
- `components/events/EventCalendarView.tsx` — calendar view variant
- `components/dashboard/UpcomingEventsCard.tsx` — dashboard card
- `app/(app)/events/page.tsx` — events listing page
- `app/(app)/events/[eventId]/page.tsx` — event detail page
- `app/(app)/groups/[slug]/page.tsx` — group hub (events tab)
- `app/(app)/groups/[slug]/GroupHubClient.tsx` — group hub client component

**Pattern — replace every bare `format()` call on event times:**

```diff
- import { format } from "date-fns";
+ import { formatInTimeZone, getTimeZoneAbbreviation } from "@/lib/utils/timezone";

- {format(startsAt, "EEE, MMM d")}
+ {formatInTimeZone(event.starts_at, viewerTimezone, "EEE, MMM d")}

- {format(startsAt, "h:mm a")} – {format(endsAt, "h:mm a")}
+ {formatInTimeZone(event.starts_at, viewerTimezone, "h:mm a")} – {formatInTimeZone(event.ends_at, viewerTimezone, "h:mm a")}
```

**Where does `viewerTimezone` come from?**
- In Server Components: fetch the viewer's profile timezone from the DB (already available in most pages via the auth check)
- In Client Components (like `EventCard`): add `viewerTimezone` as a **required** prop. Parent Server Components pass `profile.timezone`. Build fails if any parent omits it — same pattern as Step 4.

**Implementation checklist for this step:**
```bash
# Find every render of EventCard and every format() call on event dates:
grep -rn "EventCard\|format(.*startsAt\|format(.*endsAt\|format(.*starts_at\|format(.*ends_at" app/ components/
# Every EventCard render must pass viewerTimezone. Every format() on event
# dates must be replaced with formatInTimeZone(). No exceptions.
```

Since both `CreateEventForm` (Step 4) and `EventCard` (this step) use required props enforced by TypeScript, `npm run build` catches any missed render site. This is what makes the High Risk item (missing prop on parent components) a build-time error instead of a runtime bug.

**Optional enhancement** (can be a follow-up): show the event's original timezone as context:
```
Thu, Mar 26 · 11:00 AM PT (2:00 PM ET)
```

---

### Step 8 — Update recurrence generation

**File:** `lib/utils/recurrence.ts`

The current code uses UTC date math which is correct for the recurrence *interval*, but `recurrence_end_date` (a `DATE` column) should be interpreted in the event's timezone, not UTC.

**Function signature change:**

```diff
  export function buildRecurrenceDates(
    parentStartsAt: string,
    parentEndsAt: string,
    rule: "daily" | "weekly",
    recurrenceEndDate: string,
+   eventTimezone: string,
    maxInstances = rule === "daily" ? 1500 : 260,
  ): RecurrenceDatePair[] {
```

**End date interpretation change:**

```diff
- const endDate = new Date(recurrenceEndDate + "T23:59:59Z");
+ const endDate = fromZonedTime(recurrenceEndDate + "T23:59:59", eventTimezone);
```

**Callers:** Every call to `buildRecurrenceDates` must now pass the event's timezone. Grep for `buildRecurrenceDates(` to find all call sites — likely in `lib/events.ts` or an event creation action. Since this is a required parameter (not optional with a default), TypeScript catches any missed caller at build time.

This ensures "repeat until March 31" means March 31 in the event's timezone, not March 31 UTC (which could be March 30 for western US timezones).

---

### Step 9 — Update `fetchEventWithDetails` return type

**File:** `lib/events.ts`

Add `timezone` to the return type (it comes back from the DB select automatically since we're using `SELECT *`). No code change needed for the query itself — just update the TypeScript type.

Also update `lib/types.ts`:
```diff
  export interface Event {
    // ...existing fields
+   timezone: string;
  }
```

---

### Step 10 — Write tests

**File:** `lib/utils/timezone.test.ts` (new)

Test cases:
1. `localTimeToUTC` — "2:00 PM in America/New_York" → correct UTC ISO string
2. `localTimeToUTC` — DST boundary (March clock-forward, November clock-back)
3. `formatInTimeZone` — UTC string displayed correctly in ET, PT, CT
4. `getTimeZoneAbbreviation` — returns "EDT" vs "EST" depending on date
5. Round-trip: `localTimeToUTC` → `formatInTimeZone` returns original input

**Update existing:** `lib/utils/recurrence.test.ts` — add a test case where `recurrence_end_date` in a western timezone would previously have been cut short by UTC interpretation.

---

## Files Changed (Summary)

| File | Change |
|------|--------|
| `package.json` | Add `date-fns-tz` |
| `supabase/migrations/057_event_timezone.sql` | New — add `timezone` column |
| `lib/utils/timezone.ts` | New — `localTimeToUTC`, `formatInTimeZone`, `getTimeZoneAbbreviation`, `safeTimezone` |
| `lib/utils/timezone.test.ts` | New — tests for timezone utilities + invalid input handling |
| `lib/types.ts` | Add `timezone: string` to `Event` interface |
| `lib/events.ts` | Accept + pass `timezone` in `createEvent`; pass timezone to `buildRecurrenceDates` |
| `components/events/CreateEventForm.tsx` | Use `localTimeToUTC` + required `profileTimezone` prop |
| `components/events/EditEventForm.tsx` | Use event's stored TZ for form population + save |
| `components/events/EventCard.tsx` | Use `formatInTimeZone` + required `viewerTimezone` prop |
| `components/events/EventsCalendarClient.tsx` | Use `formatInTimeZone` with viewer's profile TZ |
| `components/events/EventCalendarView.tsx` | Use `formatInTimeZone` with viewer's profile TZ |
| `components/dashboard/UpcomingEventsCard.tsx` | Use `formatInTimeZone` with viewer's profile TZ |
| `app/(app)/events/page.tsx` | Pass `viewerTimezone` to `EventCard` |
| `app/(app)/events/[eventId]/page.tsx` | Use `formatInTimeZone` with viewer's profile TZ |
| `app/(app)/groups/[slug]/page.tsx` | Pass `viewerTimezone` to event components |
| `app/(app)/groups/[slug]/GroupHubClient.tsx` | Use `formatInTimeZone` with viewer's profile TZ |
| `lib/utils/recurrence.ts` | Add `eventTimezone` parameter, interpret end date in event TZ |
| `lib/utils/recurrence.test.ts` | Add timezone-aware test cases (west + east of UTC) |
| `lib/events.test.ts` | Update `createEvent` calls with `timezone` param |

---

## What this does NOT change

- **Database column types** — `starts_at` and `ends_at` stay as `TIMESTAMPTZ`. No data migration needed.
- **Existing UTC values** — All stored times remain correct. The backfill in Step 2 only sets the new `timezone` column.
- **Server-side comparisons** — `fetchUpcomingEvents` comparing against `new Date().toISOString()` is already correct (UTC vs UTC).
- **RLS policies** — No security implications. The new column is purely informational.

---

## Test Impact

**Existing tests requiring updates:**
- `lib/events.test.ts` — `createEvent` gains a `timezone` parameter; update call sites and insert payload assertions
- `lib/utils/recurrence.test.ts` — recurrence generation now takes an event timezone; update existing calls and add a western-timezone off-by-one test case

**New test file:**
- `lib/utils/timezone.test.ts` — `localTimeToUTC`, `formatInTimeZone`, `getTimeZoneAbbreviation` + DST boundary round-trips

**Unaffected (10 files):** `lib/utils.test.ts`, `lib/utils/time.test.ts`, `lib/utils/format.test.ts`, `lib/utils/status.test.ts`, `lib/utils/avatar.test.ts`, `lib/utils/events.test.ts`, `lib/utils/video.test.ts`, `lib/groups.test.ts`, `lib/messages.test.ts`, `lib/storage.test.ts`, `app/(app)/tracker/constants.test.ts`, `app/(auth)/login/login-form.test.tsx`

---

## Risks — What Can Break

### High risk (mitigated by design)

1. **`CreateEventForm` and `EventCard` parent components**
   These components need new timezone props (`profileTimezone` and `viewerTimezone`). Every parent server component must fetch and pass them. **Mitigation:** Both props are typed as required (not optional, no fallback). If any parent omits them, `npm run build` fails with a TypeScript error. Since `npm run build` is a required gate before merging (per contribution standards), this cannot ship broken. See Steps 4 and 7 for implementation checklists with grep commands to find every render site.

### Medium risk

2. **Recurrence off-by-one regression**
   Changing `"T23:59:59Z"` (hardcoded UTC) to `fromZonedTime(... eventTimezone)` fixes the western-timezone bug but could introduce a regression for events *east* of UTC (e.g., Europe/Berlin) if the math shifts the cutoff the other direction. Need to verify both directions.

3. **Missed display surfaces**
   Step 7 now lists all 8 known display surfaces found via codebase grep. However, future components added before this PR merges could introduce new `format()` calls on event dates. Run the grep in Mitigation #3 as a final check before merging.

### Low risk

4. **Existing events display at wrong times after deploy**
   The backfill sets `timezone` to the host's *current* profile timezone. If a host relocated since creating the event, the backfill assigns the wrong timezone, shifting display times. **Mitigated by the prereq PR** (`fix/profile-timezone-detection`), which ensures most users have accurate profile timezones before the backfill runs. The remaining risk is limited to hosts who genuinely relocated — a narrow edge case affecting a small number of existing events that age out quickly.

5. **Users who never set their profile timezone**
   **Mitigated by the prereq PR.** New users set timezone during onboarding. Existing users on the `America/Chicago` default get auto-corrected on next login. The only remaining gap is users who haven't logged in since the prereq shipped — they'll be corrected on their next visit, before they create any new events.

6. **Edit form timezone mismatch (UX, not a bug)**
   If a user in PST edits an event created in ET, the form correctly shows ET times (per the plan). But if they don't notice the "Times are in America/New_York" label and mentally read the times as PST, they might "correct" them and accidentally shift the event by 3 hours.

### What probably won't break

- **UTC storage is unchanged** — no data migration risk on `starts_at`/`ends_at`
- **RLS policies** — new column is informational, no security surface
- **Server-side comparisons** (`fetchUpcomingEvents`, `isLive`, `isPast`) — these compare UTC vs UTC, unaffected
- **Non-event features** — messaging, groups, polls, jobs — completely isolated

---

## How to Test

### Automated (add to test suite)

**New file — `lib/utils/timezone.test.ts`:**
- `localTimeToUTC` round-trip: `localTimeToUTC("2026-03-26", "14:00", "America/New_York")` → `formatInTimeZone(result, "America/New_York", "HH:mm")` === `"14:00"`
- DST boundaries: spring-forward (March) and fall-back (November) in US timezones
- `formatInTimeZone` — same UTC string displayed correctly in ET, PT, CT
- `getTimeZoneAbbreviation` — returns `"EDT"` vs `"EST"` depending on date
- `safeTimezone` — null returns default, empty string returns default, invalid string (e.g., `"Not/A_Zone"`) returns default, valid string passes through

**Update — `lib/utils/recurrence.test.ts`:**
- Event in `America/Los_Angeles` with `recurrence_end_date = "2026-03-31"` should include March 31, not stop at March 30
- Event in `Europe/Berlin` (east of UTC) — verify no off-by-one in the other direction

**Update — `lib/events.test.ts`:**
- `createEvent` payload includes `timezone` field

### Manual QA (critical before merge)

| Test case | Steps | Expected |
|-----------|-------|----------|
| **Basic creation** | Set profile TZ to `America/New_York`. Create event at 2:00 PM. Check DB. | `starts_at` = 18:00 UTC (or 19:00 depending on DST). `timezone` = `America/New_York` |
| **Cross-timezone display** | User A (ET) creates event at 2 PM. User B (PT) views it. | User B sees 11:00 AM |
| **Edit round-trip** | Create event → edit it → save without changes → verify times unchanged in DB | `starts_at` before and after save are identical |
| **Browser ≠ profile TZ** | Set Mac to PST. Set profile TZ to ET. Create event at 2 PM. | Stored as 18:00 UTC (ET), not 22:00 UTC (PST). Form shows "Times are in America/New_York" |
| **Recurrence boundary** | In `America/Los_Angeles`, create weekly event. Set recurrence end to March 31. | Last instance falls on or after March 31 PT, not March 30 |
| **Default timezone user** | New user (never set TZ). Create event at 2 PM. | Stored using `America/Chicago`. Form shows "Times are in America/Chicago" |
| **East-of-UTC recurrence** | In `Europe/Berlin`, create daily event. Set recurrence end to March 31. | Last instance is March 31, not April 1 |

### Smoke test for missed display surfaces

After deploying, grep the rendered pages for a known event time. If any page shows the old browser-local time instead of the profile-adjusted time, we missed a `format()` call. Specifically check:
- Dashboard `UpcomingEventsCard`
- Group events tab (`app/(app)/groups/[slug]/`)
- Any calendar view components

---

## Mitigations

1. **Safer backfill strategy** — Consider defaulting existing events to `UTC` instead of guessing from the host's current profile. Existing events would then display the same as before (browser-local interpretation of UTC). Less correct long-term, but avoids silently shifting live event times. Alternatively, only backfill *future* events and leave past events as `UTC`.

2. ~~**Make profile timezone prominent**~~ — **Addressed by prereq PR** (`fix/profile-timezone-detection`). New users set timezone during onboarding; existing users on the default get auto-corrected.

3. **Grep for every `format(` call on event dates** before the PR is final. Run:
   ```bash
   grep -rn "format(.*starts_at\|format(.*ends_at\|format(.*startsAt\|format(.*endsAt" components/ app/
   ```
   Every match must be converted to `formatInTimeZone`. Any miss = inconsistent display. All 8 known display surfaces are now listed in Step 7 — this grep is the final verification that none were missed.

4. ~~**TypeScript safety for the new prop**~~ — **Built into Steps 4 and 7.** Both `profileTimezone` and `viewerTimezone` are required props. `npm run build` catches any parent that omits them.

---

## PR Scope

Single PR. One concern: "events store and respect timezones." Includes migration, utility, form changes, display changes, and tests.

If the diff gets large, split into:
- **PR 1:** Migration + utility + types + `createEvent` (backend)
- **PR 2:** Form + display component updates (frontend)


___
# SECOND MARKDOWN PLAN
since the second part was more detailed, I tried to make one that could be robust enough to avoid degradation if there was a compaction of the context
___
# Commit 2: Event Timezone Handling — Compaction-Resistant Execution Plan

**Branch:** `31-bug-time-zone-setting-in-events-listings-urgent`
**Commit 1 already done:** `fe2c267` (profile timezone detection)
**Detailed design doc:** `plans/local/fix-event-timezone-handling.md`

---

## Why This Plan Exists

The original plan (`fix-event-timezone-handling.md`) is a detailed design doc. This file restructures it into **6 self-contained phases** optimized for context-window safety:

- Each phase leaves the codebase in a **compilable state** (`npm run build` passes)
- Each phase is **resumable from codebase state alone** — no conversation history needed
- Each phase's completion is **detectable by grep** (see Checkpoint Strategy)
- The detailed design doc contains full code snippets and rationale for reference

---

## Phase A: Foundation (no UI changes, no breaking changes)

**Goal:** Install dependency, add DB column, create utility, update types.

| Action | File | Detail |
|--------|------|--------|
| Install | `package.json` | `npm install date-fns-tz` |
| Create | `supabase/migrations/057_event_timezone.sql` | `ALTER TABLE public.events ADD COLUMN IF NOT EXISTS timezone TEXT NOT NULL DEFAULT 'America/Chicago';` + backfill from host profile timezone |
| Create | `lib/utils/timezone.ts` | Four exports: `safeTimezone(tz) → string`, `localTimeToUTC(date, time, tz) → ISO string`, `formatInTimeZone(utcISO, tz, fmt) → string`, `getTimeZoneAbbreviation(utcISO, tz) → string`. Uses `fromZonedTime` and `formatInTimeZone` from `date-fns-tz`. Parameter order: date-first, tz-second (matches `date-fns-tz`). Full code in design doc Step 3. |
| Create | `lib/utils/timezone.test.ts` | Tests: round-trip localTimeToUTC→formatInTimeZone, DST spring-forward/fall-back, abbreviation EDT vs EST, safeTimezone with null/empty/invalid/valid inputs |
| Edit | `lib/types.ts` | Add `timezone: string` to `Event` interface (~line 215) |

**Backfill SQL:**
```sql
ALTER TABLE public.events
  ADD COLUMN IF NOT EXISTS timezone TEXT NOT NULL DEFAULT 'America/Chicago';

UPDATE public.events e
SET timezone = COALESCE(
  (SELECT p.timezone FROM public.profiles p WHERE p.id = e.host_id),
  'America/Chicago'
);
```

**Verify:** `npm run build` passes. New column has DEFAULT so existing inserts still work.

**Detect completion:** `lib/utils/timezone.ts` exists AND `timezone: string` in `lib/types.ts` Event interface.

---

## Phase B: Backend Plumbing

**Goal:** Server actions and lib accept `timezone`. Forms don't send it yet — build stays green.

| Action | File | Detail |
|--------|------|--------|
| Edit | `lib/events.ts` | Add `timezone?: string` (optional) to `createEvent` data param. Include in insert: `timezone: data.timezone ?? "America/Chicago"`. |
| Edit | `app/(app)/events/actions.ts` | Add `timezone?: string` (optional with fallback) to `createEventAction` formData type. Pass `timezone: formData.timezone ?? "America/Chicago"` to `createEvent()`. Pass timezone to `buildRecurrenceDates()` calls. Add `timezone` field to recurring instance bulk inserts (~line 101). |
| Edit | `lib/utils/recurrence.ts` | Add `eventTimezone: string` param after `recurrenceEndDate`. Import `{ fromZonedTime }` from `date-fns-tz`. Change line `const endDate = new Date(recurrenceEndDate + "T23:59:59Z")` → `const endDate = fromZonedTime(recurrenceEndDate + "T23:59:59", eventTimezone)`. |
| Edit | `app/(app)/events/actions.ts` | Update `updateEventAction` similarly — accept optional `timezone`, pass through to update query. |

**IMPORTANT: `timezone` is optional in Phase B** to avoid breaking the build. Phase C makes it required.

**Verify:** `npm run build` passes. All existing code still compiles.

**Detect completion:** `timezone` appears in `createEventAction` formData type in `app/(app)/events/actions.ts`.

---

## Phase C: Create Event Form

**Goal:** Creator's profile timezone drives UTC conversion. Form sends timezone to server.

| Action | File | Detail |
|--------|------|--------|
| Edit | `app/(app)/events/create/page.tsx` | After auth check, fetch profile: `const { data: profile } = await supabase.from("profiles").select("timezone").eq("id", user.id).single()`. Pass `profileTimezone={profile?.timezone ?? "America/Chicago"}` to `<CreateEventForm>`. |
| Edit | `components/events/CreateEventForm.tsx` | (1) Add `profileTimezone: string` to `CreateEventFormProps` (required). (2) Remove `const timezone = Intl.DateTimeFormat()...`. (3) Import `{ localTimeToUTC }` from `@/lib/utils/timezone`. (4) Replace `new Date(\`${date}T${startTime}\`).toISOString()` → `localTimeToUTC(date, startTime, profileTimezone)`. Same for endTime. (5) Add `timezone: profileTimezone` to formData sent to action. (6) Update the "Times are in..." label to show `profileTimezone`. |
| Edit | `app/(app)/events/actions.ts` | Change `timezone?: string` → `timezone: string` (now required — form always sends it). Remove the `?? "America/Chicago"` fallback. |

**Verify:** `npm run build` passes. Create event → check DB: `starts_at` is correct UTC, `timezone` column matches profile.

**Detect completion:** `profileTimezone` appears as prop in `components/events/CreateEventForm.tsx`.

---

## Phase D: Edit Event Form

**Goal:** Edit form loads and saves using the event's stored timezone, not browser timezone.

| Action | File | Detail |
|--------|------|--------|
| Edit | `app/(app)/events/[eventId]/edit/page.tsx` | Add `timezone` to the select query (line 24): `"id, title, description, event_type, host_id, starts_at, ends_at, location, max_attendees, group_id, timezone"`. Pass `timezone: event.timezone` in the event prop object to `<EditEventForm>`. |
| Edit | `components/events/EditEventForm.tsx` | (1) Add `timezone: string` to the event prop type. (2) Import `{ formatInTimeZone, localTimeToUTC }` from `@/lib/utils/timezone`. (3) Replace `format(startsAt, "yyyy-MM-dd")` → `formatInTimeZone(event.starts_at, event.timezone, "yyyy-MM-dd")`. (4) Replace `format(startsAt, "HH:mm")` → `formatInTimeZone(event.starts_at, event.timezone, "HH:mm")`. (5) Same for `endsAt`/`ends_at`. (6) Replace save conversion: `new Date(\`...\`).toISOString()` → `localTimeToUTC(date, time, event.timezone)`. (7) Add `timezone: event.timezone` to the update payload. (8) Show "Times are in {event.timezone}" label near the time fields. |

**Verify:** `npm run build` passes. Edit existing event → save without changes → DB `starts_at` unchanged.

**Detect completion:** `formatInTimeZone` imported in `components/events/EditEventForm.tsx`.

---

## Phase E: Display Surfaces (8 files)

**Goal:** All event time rendering uses `formatInTimeZone` with viewer's profile timezone.

Each change is mechanical: replace `format(new Date(event.starts_at), pattern)` → `formatInTimeZone(event.starts_at, viewerTimezone, pattern)`.

**Viewer timezone fetch pattern** (used in every Server Component):
```typescript
const { data: viewerProfile } = await supabase
  .from("profiles")
  .select("timezone")
  .eq("id", user.id)
  .single();
const viewerTimezone = viewerProfile?.timezone ?? "America/Chicago";
```

### E1: EventCard (core component — 4 render sites depend on this)

**File:** `components/events/EventCard.tsx`
- Add `viewerTimezone: string` to `EventCardProps` (required)
- Import `{ formatInTimeZone }` from `@/lib/utils/timezone`
- Replace `format(startsAt, "EEE, MMM d")` → `formatInTimeZone(event.starts_at, viewerTimezone, "EEE, MMM d")`
- Replace `format(startsAt, "h:mm a")` → `formatInTimeZone(event.starts_at, viewerTimezone, "h:mm a")`
- Replace `format(endsAt, "h:mm a")` → `formatInTimeZone(event.ends_at, viewerTimezone, "h:mm a")`
- Remove `const startsAt = new Date(event.starts_at)` and `const endsAt = new Date(event.ends_at)` if no other code uses them

### E2: Events page (passes viewerTimezone to EventCard + calendar)

**File:** `app/(app)/events/page.tsx`
- Fetch viewer profile timezone (pattern above) after auth check
- Pass `viewerTimezone` to all `<EventCard>` at lines ~177 and ~212
- Pass `viewerTimezone` to `<EventsCalendarClient>` at line ~162
- Add `viewerTimezone: string` to `PastEventsSection` props and pass through

### E3: EventsCalendarClient (passes through to EventCard)

**File:** `components/events/EventsCalendarClient.tsx`
- Add `viewerTimezone: string` to props
- Pass `viewerTimezone={viewerTimezone}` to `<EventCard>` at line ~82

### E4: GroupHubClient + group page

**Files:** `app/(app)/groups/[slug]/GroupHubClient.tsx` + `app/(app)/groups/[slug]/page.tsx`
- GroupHubClient: add `viewerTimezone: string` to props, pass to `<EventCard>` at line ~439
- page.tsx: fetch viewer profile timezone, pass `viewerTimezone` to `<GroupHubClient>`

### E5: UpcomingEventsCard (Server Component — fetches its own timezone)

**File:** `components/dashboard/UpcomingEventsCard.tsx`
- Fetch viewer timezone inside the component (it already has a supabase client + user)
- Import `{ formatInTimeZone }` from `@/lib/utils/timezone`
- Replace `format(startsAt, "MMM d")` → `formatInTimeZone(e.starts_at, viewerTimezone, "MMM d")`
- Replace `format(startsAt, "h:mm a")` → `formatInTimeZone(e.starts_at, viewerTimezone, "h:mm a")`
- Remove `const startsAt = new Date(e.starts_at)` if unused

### E6: Event detail page

**File:** `app/(app)/events/[eventId]/page.tsx`
- Fetch viewer profile timezone
- Replace all `format(startsAt, ...)` calls with `formatInTimeZone(event.starts_at, viewerTimezone, ...)`
- Replace `format(endsAt, ...)` similarly

### E7: EventCalendarView (day bucketing)

**File:** `components/events/EventCalendarView.tsx`
- Add `viewerTimezone: string` as required prop
- Replace `startOfDay(new Date(e.starts_at))` bucketing with `formatInTimeZone(e.starts_at, viewerTimezone, "yyyy-MM-dd")` for bucket keys
- Parent (`EventsCalendarClient`) passes viewerTimezone through

**Verify:** `npm run build` passes. Final grep:
```bash
grep -rn "format(.*starts_at\|format(.*ends_at\|format(.*startsAt\|format(.*endsAt" components/ app/
```
Should return **0 matches** (all converted to `formatInTimeZone`).

**Detect completion:** `viewerTimezone` appears as prop in `components/events/EventCard.tsx`.

---

## Phase F: Tests + Final Verification

| Action | Detail |
|--------|--------|
| Update `lib/events.test.ts` | Add `timezone` field to all `createEvent` test calls |
| Update `lib/utils/recurrence.test.ts` | Add `eventTimezone` param to all `buildRecurrenceDates` calls. Add test: `America/Los_Angeles` event with `recurrence_end_date = "2026-03-31"` should include March 31 (not stop at March 30). Add test: `Europe/Berlin` event — no off-by-one in opposite direction. |
| Run `npm run lint` | Must pass |
| Run `npm run build` | Must pass |
| Run `npm test` | Must pass |
| Final grep | `grep -rn "format(.*starts_at\|format(.*ends_at\|format(.*startsAt\|format(.*endsAt" components/ app/` → 0 matches |

**After all phases pass:** Stage all changes and commit as Commit 2.

---

## Phase Dependencies

```
A (foundation) → B (backend) → C (create form)
                              → D (edit form)
                              → E (display surfaces)
                              → F (tests — after all others)
```

C, D, and E are independent of each other but all depend on A + B.

---

## Checkpoint Strategy

If context compaction occurs mid-implementation, determine current phase by running these checks:

| Check | Command | If TRUE → phase is done |
|-------|---------|------------------------|
| Phase A | `test -f lib/utils/timezone.ts && grep -q "timezone: string" lib/types.ts` | A complete |
| Phase B | `grep -q "timezone" app/\(app\)/events/actions.ts` (in formData type) | B complete |
| Phase C | `grep -q "profileTimezone" components/events/CreateEventForm.tsx` | C complete |
| Phase D | `grep -q "formatInTimeZone" components/events/EditEventForm.tsx` | D complete |
| Phase E | `grep -q "viewerTimezone" components/events/EventCard.tsx` | E complete |
| Phase F | Run `npm test` — all pass + final grep returns 0 | F complete |

Resume from the first incomplete phase. The design doc (`plans/local/fix-event-timezone-handling.md`) has all code snippets needed.

timchaffee:
Instead of using the users stored timezone would it be better to ask what timezone they want the event to be based upon? There could be a TZ selection on the event page that defaults to the users TZ.

tonyrosario:
> Instead of using the users stored timezone would it be better to ask what timezone they want the event to be based upon? There could be a TZ selection on the event page that defaults to the users TZ.

Great question - as I wrestled that with myself!  I would defer to anyone with more UX experience or familiarity with the project.

The reason I chose to lock-in to the user's (profile) timezone was to keep the design simple for this first pass...one less field to 'touch' on the page and consistency, one less field on the backend.

I agree with you - open up the timezone drop-down and default to user's TZ.  Let me push up a change.

timchaffee:
> > Instead of using the users stored timezone would it be better to ask what timezone they want the event to be based upon? There could be a TZ selection on the event page that defaults to the users TZ.
> 
> Great question - as I wrestled that with myself! I would defer to anyone with more UX experience or familiarity with the project.
> 
> The reason I chose to lock-in to the user's (profile) timezone was to keep the design simple for this first pass...one less field to 'touch' on the page and consistency, one less field on the backend.
> 
> I agree with you - open up the timezone drop-down and default to user's TZ. Let me push up a change.

Thinking like an event coordinator - I would want to be able to say this event is at 4PM in Central Standard Time (even if I am in a different timezone). If, for now, we want all events to be in a set timezone that makes sense. Either way we probably need to display all 3 when viewing the event:

Event database Date/Time/Timezone using the timestamptz (UTC) and timezone setting
Stored event Date/Time/Timezone converted to users profile timezone
Stored event Date/Time/Timezone converted to current browser timezone

So as a person viewing the event I would want to see it as 4PM CST with a note showing what time that would be in my browsers timezone and my profiles prefered timezone. So it would display up to 3 times if they are all different. 2 times if any 2 of the 3 match and 1 time if they are all the same. We could skip the users prefered timezone and just start off with 2 times - the Date/Time/TZ of the event and display that raw and then also the conversion to the users browsers timezone. 

We could, for ease, remove the user profile timezone and just display 2 Values if different:

Event database Date/Time/Timezone using the timestamptz (UTC) and timezone setting
Stored event Date/Time/Timezone converted to current browser timezone

If we convert to UTC on the client side does the DB then take the servers date/time setting and convert it to UTC at that time? If I am reading the docs correctly Postgres timestamptz will take a date/time/timezone and convert it to UTC automatically based on the servers time setting. 

Running SELECT NOW() on the server at 12:12PM EDT on my machine set at EDT returns 

[{"now":"2026-03-21 16:12:33.165068+00"}]

and displays as this in supabase editor:

2026-03-21 16:12:33.165068+00

Running SHOW timezone; shows UTC. 
So as long as we always have the DB set to UTC sending in UTC should work. However I think a code comment where we convert it to UTC could be useful to explain the server needs to be set to UTC for it to work. It may also depend on the WEBSERVER and/or DB Client having the correct TIMEZONE that MATCHES the DB, I have not gotten a clear picture on that.

NOTE: I have not tested with Postgres that if the server is set to NON UTC and you pass in UTC it will do the conversion but it looks that way according to the docs and if a note is added then it could just say "Only tested with Supabase/Postgres set to UTC and the webserver/db client os timezone set to <get settings??>).

Would it be better to change the 2 columns from timestamptz to timestamp so there is no auto DB conversion and we always pass in UTC and the timezone of the event? Or the actual time in the timezone and the timezone? Depending on what we do here column names are going to be important to match what we are storing. If we are only storing UTC I would rename the columns to have UTC in the name even if we leave it as timestamptz so it is more self revealing. However that might mean changing a lot of column names so I defer to delivery speed and refactor later if we desire. 

Everytime I have encountered timezone issues the variables to consider are: 
Javascript and os client timezone settings
Web Server/DB client timezone settings
Database datatype and timezone settings

When testing this we would test with the production settings for the Webserver Timezone, DB Client Timezone (if set) and DB Timezone and then change the browsers /os timezones. 

A few questions that come to mind are: 

Will the DB always be set to UTC?
Will the webserver Web Server/DB client always be set to the same specific timezone (UTC?)
We can't control the browser but we can control what we send to the API from the browser. 

Depending on the answers and the longer term requirements (keeping YAGNI in mind) what is the best course of action?

tonyrosario:
This is great feedback, thank you!  Let me review and think about this (today is booked so I'll have to circle back tomorrow!)

Having said that, I don't think we're that far off.  
- I think the storage of timezones is close (or done) but needs more review + testing.  I need to review your notes more throughly.
- I didn't consider the case of the multiple timezones when viewing an event so that is probably the biggest gap.

tonyrosario:
Screenshot 2026-03-22 at 10 08 55 PM

Here's an example.  I was able to verify the dual display and then tried add the potential triple timezone display.  I put it behind a feature toggle because I wasn't sure how it would turn out.  Toggle = `NEXT_PUBLIC_SHOW_BROWSER_TIMEZONE=true`

In the triple scenario:
1. User's profile is set to TZ A (Chicago), "in your time" in the screenshot.
2. Event is created with TZ B (Los Angeles)
3. User's browser is registering in TZ C (Denver), "local time" in the screenshot.

timchaffee:
> Screenshot 2026-03-22 at 10 08 55 PM
> Here's an example. I was able to verify the dual display and then tried add the potential triple timezone display. I put it behind a feature toggle because I wasn't sure how it would turn out. Toggle = `NEXT_PUBLIC_SHOW_BROWSER_TIMEZONE=true`
> 
> In the triple scenario:
> 
> 1. User's profile is set to TZ A (Chicago), "in your time" in the screenshot.
> 2. Event is created with TZ B (Los Angeles)
> 3. User's browser is registering in TZ C (Denver), "local time" in the screenshot.

That looks great - nice work.

tonyrosario:
Going to close this one and replace with https://github.com/What-We-Will/community-platform/pull/42

I wanted a cleaner commit history so it involved many rebases.

PR #38, opened 2026-03-20
Feature/my tools profile nudges email

marymaggee:
Profile completeness nudges + optional weekly My Tools email reminders

### Summary
Builds on My Tools / Pulsar integration with in-app guidance, optional email reminders (no automatic Pulsar calls), and documentation for how refresh/cost works.

### In-app

- Profile completeness on My Tools (0–100) from headline, bio (length), skills count, location, LinkedIn—aligned with the Pulsar payload.
- Banner when score &lt; 80 with missing fields + link to My Profile.
- Stale matches note when the last match run is ≥ 7 days old; empty state copy nudges profile enrichment.

### Email & cron

- Shared lib/email.ts (Gmail via GMAIL_USER / GMAIL_APP_PASSWORD); DM notifications in app/api/messages/route.ts refactored to use it (behavior unchanged).
- Opt-in on My Profile: “Email me a weekly reminder…” (default off).
- Migration 058_profile_my_tools_email_prefs.sql: email_my_tools_reminders, last_my_tools_reminder_sent_at.
- GET /api/cron/my-tools-reminders: secured with Authorization: Bearer ${CRON_SECRET}; 7-day cooldown per member; max 40 sends per run; only approved, onboarded, opted-in users; updates last_my_tools_reminder_sent_at after send.
- vercel.json: weekly cron (Mondays 15:00 UTC).

**Docs**

- docs/mytools-refresh.md: user-triggered Pulsar refresh, no batch auto-refresh by design, cost notes, env vars, deploy checklist.
- README.md: Gmail + cron + service role env section; My Tools row in feature table.

**Types / misc**

- Profile (+ related types) and ConversationView stub profile updated for new fields.

**Deploy notes**

- Run Supabase migration 058.
- Set CRON_SECRET, SUPABASE_SERVICE_ROLE_KEY, NEXT_PUBLIC_SITE_URL, and Gmail vars on Vercel (cron availability may depend on Vercel plan).

### DIFF
--- a/app/(app)/my-tools/MyToolsClient.tsx
+++ b/app/(app)/my-tools/MyToolsClient.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import Link from "next/link";
 import { useRouter } from "next/navigation";
 import { useState, useTransition } from "react";
 import ReactMarkdown from "react-markdown";
@@ -8,6 +9,7 @@ import { Button } from "@/components/ui/button";
 import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import { generateCareerBrief, refreshLiveMatches, saveMatchToTracker } from "./actions";
 import { WhatWeWillMatch } from "@/lib/pulsar/types";
+import type { ProfileCompleteness } from "@/lib/profile-completeness";
 
 type MatchRun = {
   request_id: string;
@@ -24,12 +26,18 @@ type CareerBrief = {
   created_at: string;
 };
 
+const STALE_MATCH_DAYS = 7;
+
 export default function MyToolsClient({
   latestMatchRun,
   latestBrief,
+  profileCompleteness,
+  matchRunAgeDays,
 }: {
   latestMatchRun: MatchRun | null;
   latestBrief: CareerBrief | null;
+  profileCompleteness: ProfileCompleteness;
+  matchRunAgeDays: number | null;
 }) {
   const router = useRouter();
   const [pendingMatch, startMatch] = useTransition();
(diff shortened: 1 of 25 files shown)


app/(app)/my-tools/actions.ts
```diff
@@ -0,0 +1,155 @@
...
+    return { ok: true };
+  } catch (error) {
+    return {
+      error:
+        error instanceof Error ? error.message : "Unable to generate career brief",
+    };
+  }
+}
+
+export async function saveMatchToTracker(match: WhatWeWillMatch): Promise<ActionResult> {
```
tonyrosario, line 104: Since this is an exposed endpoint, we should do a quick security pass before merge to check for things like injection/XSS.

It currently accepts the full match object from the client and writes fields directly to `job_applications`. An attacker can send arbitrary data.

Since the match data already exists server-side in `member_match_runs.matches`, suggest changing the API to accept (matchRunId, matchIndex) and looking up the data from the user's stored run.

Also validate that url only allows https:// or http:// before storage.

kaizengrowth: It may be good to also put a unique constraint or upsert here, in the case of duplicate clicks or repeated matches, so that it doesn't create duplicate `job_applications` rows. @marymaggee

app/(app)/my-tools/MyToolsClient.tsx
```diff
@@ -0,0 +1,238 @@
...
+        setError(result.error);
+        return;
+      }
+      router.refresh();
+    });
+  }
+
+  async function onSaveToTracker(match: WhatWeWillMatch) {
+    setError(null);
+    const key = `${match.company}-${match.roleTitle}-${match.applyUrl}`;
```
kaizengrowth, line 74: NIT (low priority) - The key is `company-roleTitle-applyUrl`. Collisions are unlikely but maybe possible? What happens if there are two distinct listings with identical triple?

app/(app)/my-tools/MyToolsClient.tsx
```diff
@@ -0,0 +1,238 @@
...
+                    enrich your profile
+                  </Link>
+                  .
+                </>
+              ) : null}
+            </p>
+          ) : (
+            <>
+              <p className="text-sm text-muted-foreground">
+                Last run: {new Date(latestMatchRun.created_at).toLocaleString()}
```
kaizengrowth, line 147: I think elsewhere in the repo, we are using UTC-centric, explicity locale/time zone rather than `toLocaleString()` for the viewer's locale/device TZ.

This is okay, but I wonder if we should standardize for consistency!

@marymaggee @tonyrosario

app/(app)/my-tools/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  if (!user) {
+    redirect("/login");
+  }
+
+  const [
+    { data: latestMatchRun },
+    { data: latestBrief },
+    { data: profileRow },
+  ] = await Promise.all([
```
kaizengrowth, line 36: Nice use of parallel queries here in a clear and efficient way!

app/api/cron/my-tools-reminders/route.ts
```diff
@@ -0,0 +1,128 @@
...
+      linkedin_url: row.linkedin_url,
+      location: row.location,
+    });
+
+    const tips =
+      completeness.score < 80
+        ? completeness.missing.map((m) => missingFieldToReminderTip(m))
+        : [];
+
+    const result = await sendMyToolsReminderEmail({
```
kaizengrowth, line 93: Here an email is sent --> then update `last_my_tools_reminder_sent_at`. 

If update fails after a successful send, cooldown isn’t advanced, so the same user could be eligible again on the next cron. 

Would it be good to update first with a “pending” sentinel, or transactional outbox pattern, depending on how strict we want “at most one per 7 days” to be?

kaizengrowth (approved):
This looks awesome! Thank you for adding a "/my-tools" area with server actions to  fetch latest job match and career brief from Supbase.

marymaggee:
Superseded by unified branch feature/pulsar-my-tools-unified-pr37-pr38. Closing PR.

PR #37, opened 2026-03-20
Add My Tools Pulsar integration (matches, brief, persistence, migration)

marymaggee:
### What We Will platform + Pulsar integration plan

- Live matches (ATS) — factual, time-bound, good for “apply now” and links that should work when postings exist. This maps cleanly to the existing Pulsar contract in [docs/integrations-whatwewill.md](https://github.com/GaiaCommons/pulsar/compare/feature/docs/integrations-whatwewill.md) (POST /api/integrations/whatwewill/match).
- Broader career brief (LLM) — strategic, can include inactive or illustrative company/role examples with clear disclaimers. Keeping it separate from ATS results avoids users confusing “open role” with “directional lead.”
- My Tools — natural home for personal outputs (saved brief, last match run, links to tracker).
- Job Board + Job Application Tracker — natural home for curated postings the org trusts and state (applied, interviewed, etc.).

I'll host the LLM brief on Pulsar (second integration endpoint), so the platform stays a thin client + storage for member-facing UI. (For now I'm hosting this on marygfisher.com but will probably create a new domain at some point).

### **Phase 1 — Profile to live matches (platform + existing Pulsar)**

**Goal:** When a member (or staff) triggers “Refresh matches,” the platform calls Pulsar with a payload derived from the member profile.

**Pulsar (already exists): [src/app/api/integrations/whatwewill/match/route.ts]**(https://github.com/GaiaCommons/pulsar/compare/feature/src/app/api/integrations/whatwewill/match/route.ts) — ensure staging/production env has WHATWEWILL_API_KEY and ATS board lists.

**Community platform (github.com/What-We-Will/community-platform):**

- Add server-only config: `PULSAR_BASE_URL`, `WHATWEWILL_API_KEY` (GitHub Actions / hosting secrets — never client-side).
**- Implement a server route or server action that:**
- [ ] Loads the member profile (LinkedIn URL, resume text or parsed text, skills, industries, titles, work type, locations).
- [ ] Maps fields to WhatWeWillProfileRequest (same shape as [docs/integrations-whatwewill.md](https://github.com/GaiaCommons/pulsar/compare/feature/docs/integrations-whatwewill.md)).
- [ ] Calls POST {PULSAR_BASE_URL}/api/integrations/whatwewill/match with x-api-key.
- My Tools (v1 UI): section e.g. “Job matches (live postings)” showing matches[] (title, company, score, apply link, reasons). Optional: store last requestId + timestamp in your DB for audit.

[...]

PR #44, opened 2026-03-25
Pulsar job matches integration my tools unified pr37 pr38

marymaggee:
New unified branch feature/pulsar-my-tools-unified-pr37-pr38 based on feature/pulsar-my-tools-integration and then included the two commits from feature/my-tools-profile-nudges-email.

What’s in it:

- 21eb23e My Tools integration
- c1e47cd My Tools + jobs refactor/tests
- 61ec1f4 profile nudges + weekly email opt-in + cron + docs
- c10e20b reminder/completeness Jest tests

Overlap status
- My Tools/jobs overlap merged cleanly (auto-merge in app/(app)/my-tools/actions.ts happened during cherry-pick).
- Full test suite passes on unified branch: 22 suites, 176 tests.

This PR supersedes PR #37 and #38. 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1213823017520962


marymaggee:
Thanks @tonyrosario & @kaizengrowth for the review on the other PR. I now cleaned this up and addressed feedback items. Closing the other PRs.

**Client tampering risk fixed**
- `saveMatchToTracker` now accepts (`matchRunId`, `matchIndex`) only.
- Server loads match from `member_match_runs` for the authenticated user and validates shape before insert.
- No longer trusts full match payload from client.

**URL validation added**
- Tracker URL now normalized to only http:// or https://.
- Invalid/non-http schemes (e.g. javascript:) are stored as empty string.

**Duplicate tracker rows mitigated**
- Added migration: supabase/migrations/059_job_applications_dedupe_for_pulsar.sql
- Unique index on (`user_id`, `company`, `position`, `url`).
- Insert path switched to `upsert(..., { onConflict: "user_id,company,position,url", ignoreDuplicates: true })`.

**NIT key collision improved**
- My Tools save button key now uses stable server identifiers: matchRunId-matchIndex instead of company-roleTitle-applyUrl.

**Timestamp consistency improved**
- Replaced `toLocaleString()` in My Tools with explicit UTC formatter (`Intl.DateTimeFormat(...`, timeZone: "UTC")).

**Cron send/update ordering improved**
- Reminder cron now reserves cooldown timestamp before sending using compare-and-set semantics.
- If email send fails, performs best-effort rollback.
- This reduces repeat sends when update-after-send fails or concurrent runs overlap.

app/(app)/my-tools/MyToolsClient.tsx
```diff
@@ -0,0 +1,244 @@
...
+    setError(null);
+    const key = `${matchRunId}-${matchIndex}`;
+    setSavingIds((prev) => ({ ...prev, [key]: true }));
+    const result = await saveMatchToTracker(matchRunId, matchIndex);
+    setSavingIds((prev) => ({ ...prev, [key]: false }));
+    if (result.error) {
+      setError(result.error);
+      return;
+    }
+    router.refresh();
```
tonyrosario: I saw a warning that if there's an exception thrown, the button could be permanently disabled.  Assuming it's a valid issue, it was suggested to add a finally clause to always reset it.

```
...
    setSavingIds((prev) => ({ ...prev, [key]: true }));
    try {
      const result = await saveMatchToTracker(matchRunId, matchIndex);
      ...
      router.refresh();
    } catch (err) {
      setError(err instanceof Error ? err.message : "<some error message>");
    } finally {
      setSavingIds((prev) => ({ ...prev, [key]: false })); // always resets to false
    }
```

app/(app)/my-tools/actions.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+}
+
+export async function generateCareerBrief(): Promise<ActionResult> {
+  const supabase = await createClient();
+  const {
+    data: { user },
+  } = await supabase.auth.getUser();
+  if (!user) return { error: "Not authenticated" };
+
```
tonyrosario, line 98: A potential issue was found by code reviewer.  Would ask that someone review this as I'm not familiar enough with Pulsar.   👀 

[WARNING] `app/(app)/my-tools/actions.ts` — no per-user rate limiting on Pulsar calls

Problem: `refreshLiveMatches` and `generateCareerBrief` call external Pulsar APIs (ATS + LLM) with no per-user rate limiting.

> Any authenticated member can call these server actions in a tight loop, generating unbounded LLM and ATS API cost.

Fix: Before calling Pulsar, query the user's most recent run/brief timestamp and return an error if within a minimum cooldown window (e.g., 5 minutes).

app/api/cron/my-tools-reminders/route.ts
```diff
@@ -0,0 +1,148 @@
...
+    )
+    .slice(0, MAX_SEND_PER_RUN);
+
+  let sent = 0;
+  let failed = 0;
+  const errors: string[] = [];
+
+  for (const row of eligible) {
+    const { data: authUser, error: authErr } =
+      await service.auth.admin.getUserById(row.id);
```
tonyrosario: Another issue flagged by code reviewer that I think needs human review 👀 .

> Performance concern: serial Auth Admin API calls in loop

>`service.auth.admin.getUserById(row.id)` is called once per eligible user inside a for loop — up to 40 sequential network round-trips to the Supabase Auth Admin API per cron invocation. This adds latency and puts us at risk of hitting Auth Admin rate limits under load, with no retry path when it fails.

>A few options to consider:
> - Store email on profiles — sync from auth.users via trigger, read it in the existing query, no loop needed
> - Parallelize with Promise.all — reduces wall-clock time but doesn't eliminate the Auth Admin calls or the rate limit risk
> - Batch lookup — check if the Auth Admin API supports a bulk endpoint; if so, one call replaces the loop

tonyrosario:
FYI - this branch may need to migrate to the new timestamp format for migrations due to docs/adr/0002.

[Some info in my comment on PR 128](https://github.com/What-We-Will/community-platform/pull/128#issuecomment-4512732557).

Just let me know if you need help!

PR #56, opened 2026-03-31
Update preview YAML for Supabase DB Branching

timchaffee:
Deleted


timchaffee:
Supabase does not support bracnhing unless you get pro - going to close this PR without merging

PR #61, opened 2026-04-02
feat: add sticky styling to profile, logout section in app-shell

GeorgeDover:
Asana: https://app.asana.com/1/1213515485882011/project/1213516055535837/task/1213904250386259?focus=true

UI/UX improvement on Dashboard and other pages where left nav is present from app-shell. 

Currently profile icon + details, bug reporting and logout buttons are all fixed at the extreme bottom of the left nav pane. 

Would be better UX (at least for now while we don't have too many nav options) to stick it to the bottom of the screen within the nav pane.

Before:
image

After:
image

### DIFF
--- a/app/(app)/app-shell.tsx
+++ b/app/(app)/app-shell.tsx
@@ -252,7 +252,7 @@ export default function AppShell({ children, user }: AppShellProps) {
             )}
           </nav>
 
-          <div className="border-t p-4">
+          <div className="sticky bottom-0 z-30 border-t bg-background p-4">
             <div className="flex items-center gap-3 rounded-lg px-3 py-2">
               <UserAvatar
                 avatarUrl={user.avatarUrl}


tonyrosario:
Thanks @GeorgeDover!

After checking on my local, the profile is sticky but it's not part of the nav bar - was that the intent?

What I'm seeing:
Screenshot 2026-04-03 at 10 23 06 AM

What I thought was the intent:
(and maybe I'm reading the description incorrectly)

Screenshot 2026-04-03 at 10 24 30 AM

kaizengrowth (approved):
This is a small change that makes a

GeorgeDover:
> Thanks @GeorgeDover!
> 
> After checking on my local, the profile is sticky but it's not part of the nav bar - was that the intent?
> 
> What I'm seeing: Screenshot 2026-04-03 at 10 23 06 AM
> 
> What I thought was the intent: (and maybe I'm reading the description incorrectly)
> Screenshot 2026-04-03 at 10 24 30 AM

Sorry missed this comment! Yes this is the intended behavior, but maybe I didn't explain the intent properly. Basically I wanted the profile, bug reporting and logout section to be at the bottom left of the screen no matter where you are scroll-wise. We could adapt it so that it immediately follows on from the other nav options, but this won't be a good experience if you have to scroll a long way on a very long page (the logout will be somewhere up above the viewport). My solution as well is not great for mobile/magnification/lots of navigation links above, but we will have to rethink the whole UI/UX at those points anyway so I felt this was a good stop-gap :)

PR #52, opened 2026-03-30
Fix/31 timezone fast follow pr feedback v2

tonyrosario:
recreation of https://github.com/What-We-Will/community-platform/pull/46 because Vercel build issues.

> @tonyrosario I think this branch was started from an earlier main that didn't contain the github workflows or the vercel settings. I suspect / hope a rebase may resolve.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1213746764285240

PR #62, opened 2026-04-02
Messagetrash

### LINKED ISSUE #123: Unable to leave group as the sole member
## Description
When a user is the only member of a group, the "Leave" button on the group learning page does not work.

## How to Reproduce
1. Create or join a group where you are the only member
2. Click the leave button on the right-hand side of the group page
3. Nothing happens — you remain in the group

## Expected Behavior
The sole member should be able to leave the group. The group could be auto-archived when the last member leaves.

## Relevant Files
- `app/(app)/groups/actions.ts` — `leaveGroupAction()` (~line 79-99), may have a guard preventing the last member from leaving

## Fix
Check the edge case in `leaveGroupAction` — if it blocks the last member from leaving, either remove that guard or auto-archive the group when the last member leaves.

OmegaDeanGrey:
Removed Group Chats from ConversationsList so they could not be deleted entirely by a single participant, added delete function and button with trash icon, corrected incorrect icon spelling sendHorizonal to sendHorizontal from MessageInput.tsx

MockConversationList component created for testing ConversationList.
Includes DM, group conversations, with proper typing fixes for Message.metadata and Conversation fields.

Run `npm run dev` and navigate to `/ConversationList_test/page` to see the mocked ConversationList.
Group conversations are included but filtered out in the component.

Resolves inability to delete conversations, prevents group chat from being affected or cluttering up DMs, adds the trash icon next to the archive button on CoversationList.

Next steps are to display archived messages, ensure double loading messages have a sending state, and group chats can be put in separate panel later

Closes #123
Group conversations are accessible still through groups

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -28,3 +28,8 @@ ACTION_NETWORK_API_KEY=
 # Server-side only — never expose as NEXT_PUBLIC_.
 GOOGLE_FORM_SHARE_YOUR_STORY_URL=
 
+
+# .env.local
+ACTION_NETWORK_API_KEY=smurf
+SUPABASE_URL=smurf
+SUPABASE_ANON_KEY=smurf
--- a/app/ConversationList_test/page.tsx
+++ b/app/ConversationList_test/page.tsx
@@ -1,0 +1,168 @@
+"use client";
+
+import { ConversationList } from "@/components/messages/ConversationList";
+import type { ConversationWithDetails, Message, Profile } from "@/lib/types";
+
+// Mock utility functions
+const formatRelativeTime = (date: string) =>
+  new Date(date).toLocaleTimeString();
+const getAvatarColor = (_name: string) => "bg-blue-500 text-white";
+const LiveStatusAvatar = ({
+  avatarUrl,
+  displayName,
+}: {
+  avatarUrl: string | null;
+  displayName: string;
+}) => (
+  <div className="w-8 h-8 rounded-full bg-gray-300 flex items-center justify-center text-xs">
+    {displayName[0]}
+  </div>
+);
+
+export default function MockConversationList() {
+  const currentUserId = "user_1";
+  const selfNotesId = "notes_1";
+
+  const initialConversations: ConversationWithDetails[] = [
+    // My Notes
+    {
+      conversation: {
+        id: selfNotesId,
+        type: "dm",
+        group_id: null,
+        created_at: new Date().toISOString(),
+      },
+      participants: [],
+      lastMessage: {
+        id: "msg_0",
+        conversation_id: selfNotesId,
+        sender_id: currentUserId,
+        content: "Remember to mock this component!",
+        created_at: new Date().toISOString(),
+        message_type: "system",
+        metadata: {},
+        edited_at: null,
+      },
+      unreadCount: 0,
+    },
+    // DM with Alice
+    {
+      conversation: {
+        id: "conv_2",
+        type: "dm",
+        group_id: null,
+        created_at: new Date().toISOString(),
+      },
+      participants: [
+        {
+          id: "user_2",
+          display_name: "Alice",
+          avatar_url: null,
+          resume_path: null,
+          headline: null,
+          bio: null,
+          location: null,
+          skills: [],
+          open_to_referrals: true,
+          linkedin_url: null,
+          github_url: null,
+          portfolio_url: null,
+          timezone: "UTC",
+          is_onboarded: true,
+          approval_status: "approved",
+          role: "member",
+          last_seen_at: new Date().toISOString(),
+          created_at: new Date().toISOString(),
+          updated_at: new Date().toISOString(),
+        },
+      ],
+      lastMessage: {
+        id: "msg_1",
+        conversation_id: "conv_2",
+        sender_id: "user_2",
+        content: "Hey there!",
+        created_at: new Date().toISOString(),
+        message_type: "text",
+        metadata: {},
+        edited_at: null,
+      },
+      unreadCount: 1,
+    },
+    // DM with Bob
+    {
+      conversation: {
+        id: "conv_3",
+        type: "dm",
+        group_id: null,
+        created_at: new Date().toISOString(),
+      },
+      participants: [
+        {
+          id: "user_3",
+          display_name: "Bob",
+          avatar_url: null,
+          resume_path: null,
+          headline: null,
+          bio: null,
+          location: null,
+          skills: [],
+          open_to_referrals: true,
+          linkedin_url: null,
+          github_url: null,
+          portfolio_url: null,
+          timezone: "UTC",
+          is_onboarded: true,
+          approval_status: "approved",
+          role: "member",
+          last_seen_at: new Date().toISOString(),
+          created_at: new Date().toISOString(),
+          updated_at: new Date().toISOString(),
+        },
+      ],
+      lastMessage: {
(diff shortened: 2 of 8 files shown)


kaizengrowth:
Awesome work on this, @OmegaDeanGrey! I'll review this code today. 

Would appreciate your thoughts on this as well @tonyrosario. Thank you!

test.txt
```diff
@@ -0,0 +1 @@
+// test change
```
tonyrosario, line 1: Note - when the PR is finalized, this commit should be deleted via interactive rebase.

components/messages/ConversationList.tsx
```diff
@@ -130,7 +134,29 @@ export function ConversationList({
...
+
+  async function handleDelete(
+    e: React.MouseEvent<HTMLButtonElement>,
+    conversationId: string,
+  ) {
+    e.preventDefault();
+    e.stopPropagation();
+
+    const supabase = createClient();
+    await supabase.from("conversations").delete().eq("id", conversationId);
```
tonyrosario, line 153: Few issues identified that need to be addressed:

1. **Blocker:** no auth/participation check (any authenticated user can attempt to delete any conversationId)

2. **Blocker:** since there isn't a RLS DELETE policy on conversations, the delete silently fails at the DB level

- ❓ What should the policy be?  @kaizengrowth @SirFuzzyWalls 
  1. only allow delete if the user is the only remaining participant?
  2. restrict it to the conversation creator?
  3. something else...?

3. result is never inspected so the UI removes the row regardless of success/failure.

suggest:
1. Move this to a server action that calls `supabase.auth.getUser()`
2. Check if they are a valid participant (conversation_participants) for the conversationId:
  💹 if yes, return the result to client and then UI updates on success
  ❌ if no, fail fast with an error

You can add `deleteConversation` as a function in lib/actions/messages.ts

components/messages/ConversationList.tsx
```diff
@@ -285,12 +438,14 @@ export function ConversationList({
                     </Link>
                     <TooltipProvider>
                       <Tooltip>
-                        <TooltipTrigger asChild>
+                        <TooltipTrigger
+                          asChild
+                          onClick={(e) => handleArchive(e, conversation.id)}
```
tonyrosario: probably worth calling out that `handleArchive` has the same problem as `handleDelete` but we can add a separate issue for that.

.env.example
```diff
@@ -28,3 +28,8 @@ ACTION_NETWORK_API_KEY=
 # Server-side only — never expose as NEXT_PUBLIC_.
 GOOGLE_FORM_SHARE_YOUR_STORY_URL=
 
+
+# .env.local
+ACTION_NETWORK_API_KEY=smurf
+SUPABASE_URL=smurf
+SUPABASE_ANON_KEY=smurf
```
tonyrosario, line 35: this reminds me that I need to get https://github.com/What-We-Will/community-platform/pull/52 merged.

That should address these (and can be removed)

package.json
```diff
@@ -51,6 +51,6 @@
     "supabase": "^2.76.15",
     "tailwindcss": "^4",
     "tw-animate-css": "^1.4.0",
-    "typescript": "^5"
+    "typescript": "^5.9.3"
```
tonyrosario, line 54: Does this feature depend on updating the pins on these dependencies?  If not, we should roll these back.

Dependency changes should come in their own PR with a reason for the change

components/dashboard/ErrorCard.tsx
```diff
@@ -0,0 +1,22 @@
+type ErrorCardProps = {
+  message: string | null; // can be string or null
+};
+
+export default function ErrorCard({ message }: ErrorCardProps) {
```
tonyrosario, line 5: is this card necessary?  I don't think it's being used anywhere (after looking at the full commit)

app/ConversationList_test/page.tsx
```diff
@@ -0,0 +1,168 @@
...
+}: {
+  avatarUrl: string | null;
+  displayName: string;
+}) => (
+  <div className="w-8 h-8 rounded-full bg-gray-300 flex items-center justify-center text-xs">
+    {displayName[0]}
+  </div>
+);
+
+export default function MockConversationList() {
```
tonyrosario, line 22: remove test page in final version.

components/messages/MessageInput.tsx
```diff
@@ -43,20 +46,36 @@ export function MessageInput({
...
 
+  async function send() {
+    if (sending) return;
+    if (!value.trim() && !attachment) return;
+    if (disabled) return;
+
+    setSending(true);
+
+    try {
+      console.log("SEND CALLED"); // temp debug
```
tonyrosario, line 57: remove temp debug statement before merging

components/messages/MessageInput.tsx
```diff
@@ -43,20 +46,36 @@ export function MessageInput({
...
+
+      await onSend(value.trim(), attachment ?? undefined);
+
+      // clear AFTER send
+      setAttachment(null);
+      setAttachmentError(null);
+      onChange("");
```
tonyrosario, line 64: NIT: Avoid unnecessary comments, keep the code self-documenting.

Since cleanup instructions are related, these could be extracted into a function to keep them coupled and improve maintainability (e.g. over time, people start adding code in-between these statements or delete them.

```
function clearInput() {
    setAttachment(null);
    setAttachmentError(null);
    onChange("");
}
```

tonyrosario (changes requested):

components/messages/MessageInput.tsx
SirFuzzyWalls: Agreed, good call on cleaning up after the send @OmegaDeanGrey, this is a great place to write a function so we can improve the developer experience going forward. 

As @tonyrosario also pointed out, sometimes a function name is enough clarity so comments aren't needed. 

Thanks for this work, I look forward to seeing the revised version!

OmegaDeanGrey:
Not quite sure what you mean:
NIT: Avoid unnecessary comments, keep the code self-documenting.

Since cleanup instructions are related, these could be extracted into a function to keep them coupled and improve maintainability (e.g. over time, people start adding code in-between these statements or delete them.

function clearInput() {
    setAttachment(null);
    setAttachmentError(null);
    onChange("”);

PLEASE ADVISE

SirFuzzyWalls:
@OmegaDeanGrey here's a brief wiki article discussing the idea: 
[wiki: self documenting code](https://en.wikipedia.org/wiki/Self-documenting_code)

I'd also suggest checking this out as this is a good example of both ideas: 
[wiki: dont repeat yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

In my experience, self documenting code is a worthy goal but can be difficult to achieve 100% of the time in larger projects. In this case, these three steps are likely going to be repeated over and over again. We can either repeat these steps in multiple places or pull them out into a function that can be called when needed. 

The function will allow you to drop the comment, because it's simple enough that the function name can and should explain exactly what it does. Additionally it will allow anyone else to easily make changes if needed and see them reflected in multiple places.

So this: 

```
// clear AFTER send
      setAttachment(null);
      setAttachmentError(null);
      onChange("");
```
      

Becomes: 
```
function clearInput() {
    setAttachment(null);
    setAttachmentError(null);
    onChange("");
}
```

OmegaDeanGrey:
Ok thank you

PR #67, opened 2026-04-04
update route to allow update-password

### LINKED ISSUE #66: Forgot password does not redirect to updatepassword
When using the forgot password flow it lands on the dashboard page instead of the update-password page. It should land on the update password page so the user can reset thier password.

timchaffee:
Update proxy to ignore update-password with code querystring

### DIFF
--- a/proxy.ts
+++ b/proxy.ts
@@ -55,8 +55,7 @@ export async function proxy(request: NextRequest) {
   if (
     code &&
     pathname !== "/auth/callback" &&
-    !pathname.startsWith("/auth/callback/") &&
-    !pathname.startsWith("/update-password") 
+    !pathname.startsWith("/auth/callback/")
   ) {
     const url = request.nextUrl.clone();
     url.pathname = "/auth/callback";
--- a/proxy.ts
+++ b/proxy.ts
@@ -55,7 +55,8 @@ export async function proxy(request: NextRequest) {
   if (
     code &&
     pathname !== "/auth/callback" &&
-    !pathname.startsWith("/auth/callback/")
+    !pathname.startsWith("/auth/callback/") &&
+    !pathname.startsWith("/update-password") 
   ) {
     const url = request.nextUrl.clone();
     url.pathname = "/auth/callback";
--- a/app/(auth)/update-password/update-password-form.tsx
+++ b/app/(auth)/update-password/update-password-form.tsx
@@ -26,13 +26,7 @@ export function UpdatePasswordForm({
   const [error, setError] = useState<string | null>(null);
   const [isLoading, setIsLoading] = useState(false);
   const router = useRouter();
-  const searchParams = useSearchParams();
-  const code = searchParams.get('code')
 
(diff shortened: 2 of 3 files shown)


SirFuzzyWalls (dismissed):
LGTM! Nice find.

proxy.ts
```diff
@@ -55,7 +55,8 @@ export async function proxy(request: NextRequest) {
   if (
     code &&
     pathname !== "/auth/callback" &&
-    !pathname.startsWith("/auth/callback/")
+    !pathname.startsWith("/auth/callback/") &&
+    !pathname.startsWith("/update-password")
```
tonyrosario: Security review agent found that this bypass breaks the PKCE exchange.  TIL!

I made a [branch](https://github.com/What-We-Will/community-platform/compare/66-forgot-password-does-not-redirect-to-updatepassword...66-forgot-password-does-not-redirect-to-updatepassword-pkce-fix?expand=1) just to see what changes would be involved (as that might be easier than the report output, still working on that!)

* * * 

**Why the proxy bypass doesn't work**

The proxy redirects all ?code= requests to /auth/callback because Supabase uses PKCE (Proof Key for Code Exchange) for password resets, OAuth, and magic links. The code in the URL is a one-time token that must be exchanged server-side for a session before any authenticated action can happen.

Bypassing the redirect for /update-password means the code arrives at the form but is never exchanged. The form calls supabase.auth.updateUser({ password }), which requires an active session — so the update silently fails.

The real problem is that /auth/callback successfully exchanges the code but doesn't know where the user was trying to go, so it defaults to /dashboard.

Reference: Supabase Auth - [Redirect URLs and Wildcards](https://supabase.com/docs/guides/auth/redirect-urls) and [Server-side Auth with PKCE flow](https://supabase.com/docs/guides/auth/server-side/creating-a-client?queryGroups=framework&router=app&framework=nextjs)

**How to fix it**

1. Revert the proxy.ts change. The redirect to /auth/callback is correct and must not be bypassed for any route.
2. Pass the original path through the redirect. When the proxy redirects `?code=` requests to `/auth/callback`, it should include the original pathname as a next query parameter so the callback knows where to send the user after exchange.
3. Remove the early-return in the callback route. There's a block that checks for `next=/update-password` and redirects _without exchanging the code_ — this needs to go. The callback already has logic that respects a next parameter after a successful exchange; let that handle it.
4. Remove the code param logic from the update-password form. The form no longer receives a `?code=param`. By the time the user reaches `/update-password`, the callback has already exchanged the code and set session cookies. The form just needs to call `updateUser()`.

timchaffee: Okay - I think I'll need to review what's in app/auth/callback/route.ts as it may be the issue

timchaffee: Good catch Tony - that was my misunderstanding of how the forgot/update PW worked with supabase. Should be fixed now.

SirFuzzyWalls (approved):
👍

PR #69, opened 2026-04-07
Add pre-commit hooks and workflow enhancement

SirFuzzyWalls:
Sets up Husky and lint-staged for local dev, adds type checking to both CI/CD workflows, and introduces non-blocking dependency auditing on PRs.

### Testing note:

Because this branch adds two new dependencies, you'll need to run `npm install` before you see any hooks or changes take effect. 

### Changes
- **Pre-commit**: lint-staged runs ESLint on staged `.ts`/`.tsx` files only
- **Pre-push**: `tsc --noEmit` catches type errors before they hit CI
- **`preview.yml` (PR)**: added type check step and `npm audit --audit-level=high` (non-blocking)
- **`production.yml` (Production)**: added type check step
- Fixed pre-existing type errors in `get-site-url.test.ts` surfaced by the new hook

### Why
Keeps `main` stable without adding friction for contributors. Local hooks give fast feedback without being painfully slow. The PR workflow is the strongest gate.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -89,23 +89,16 @@ jobs:
 
       - name: Comment PR with URL
         uses: actions/github-script@v7
+        env:
+          PREVIEW_URL: ${{ steps.deploy.outputs.preview_url }}
         with:
           script: |
-            const branch = context.ref.replace('refs/heads/', '');
-            const { data: prs } = await github.rest.pulls.list({
+            await github.rest.issues.createComment({
+              issue_number: context.payload.pull_request.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              head: `${context.repo.owner}:${branch}`,
-              state: 'open',
+              body: `✅ Preview deployed: ${process.env.PREVIEW_URL}`,
             });
-            for (const pr of prs) {
-              await github.rest.issues.createComment({
-                issue_number: pr.number,
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                body: '✅ Preview deployed: ${{ steps.deploy.outputs.preview_url }}'
-              });
-            }
 
       - name: Slack notify success
         if: success()
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -7,6 +7,7 @@ on:
 env:
   VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
   VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+  HUSKY: 0
 
 permissions:
   contents: read
@@ -35,9 +36,16 @@ jobs:
 
       - name: Lint
         run: npm run lint
+      
+      - name: Type check
+        run: npx tsc --noEmit
 
       - name: Run unit tests
         run: npm test
+
+      - name: Audit dependencies
+        run: npm audit --audit-level=high
+        continue-on-error: true
 
       - name: Setup Supabase CLI
         uses: supabase/setup-cli@v1
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -88,16 +88,26 @@ jobs:
           echo "preview_url=$url" >> $GITHUB_OUTPUT
 
       - name: Comment PR with URL
+        if: github.event_name == 'pull_request'
         uses: actions/github-script@v7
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -45,12 +46,17 @@
...
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
+    "lint-staged": "^16.4.0",
     "shadcn": "^3.8.5",
     "supabase": "^2.76.15",
     "tailwindcss": "^4",
     "tw-animate-css": "^1.4.0",
     "typescript": "^5"
+  },
+    "lint-staged": {
```
tonyrosario: NIT: formatting (indentation)

.github/workflows/preview.yml
```diff
@@ -35,10 +36,17 @@ jobs:
...
+      
+      - name: Type check
+        run: npx tsc --noEmit
 
       - name: Run unit tests
         run: npm test
 
+      - name: Audit dependencies
+        run: npm audit --audit-level=high
+        continue-on-error: true
```
tonyrosario, line 48: sanity check - do we want this in production.yml as well?

yes it would be redundant and scenarios it would cover would be rare, something like:
- vulnerability is published between the PR merge and prod deploy
- hotfix, where we deploy to prod and skip staging

any others?

.github/workflows/preview.yml
```diff
@@ -89,23 +89,16 @@ jobs:
...
         with:
           script: |
-            const branch = context.ref.replace('refs/heads/', '');
-            const { data: prs } = await github.rest.pulls.list({
+            await github.rest.issues.createComment({
+              issue_number: context.payload.pull_request.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              head: `${context.repo.owner}:${branch}`,
-              state: 'open',
```
tonyrosario, line 91: Ran a quick review on the last commit.


Three findings on .github/workflows/preview.yml:

  1. Line 96 — context.payload.pull_request is undefined on non-PR triggers
  Accessing .number will throw if this workflow fires on a push event. Safest
  fix is a step-level condition on the "Comment PR with URL" step:
```
  if: github.event_name == 'pull_request'
```

  2. Line 99 — Empty preview URL posts a misleading comment
  If the deploy step fails and preview_url is unset, the comment reads "✅
  Preview deployed: " with a blank URL. Add an early return before
  createComment:
```
  if (!process.env.PREVIEW_URL) return;
```
  3. Line 99 — Preview URL is unvalidated before posting
  PREVIEW_URL is sourced from deploy output and embedded directly in the
  comment body. If Vercel embeds the branch name in the preview URL, a
  maliciously named branch could post a deceptive link. Assert it starts with
  your known Vercel preview domain before posting.
```
const url = process.env.PREVIEW_URL;
  if (!url || !url.startsWith('https://your-project.vercel.app')) return;
```

--- AI-assisted review (AI_POLICY.md). Validate before treating as blockers.

tonyrosario (commented):
~~LGTM - no blockers~~

The last commit seemed to spawn a couple of issues

timchaffee (dismissed):
LGTM - please review Tony's comments

.github/workflows/preview.yml
SirFuzzyWalls: Those are the only two I can think of, currently this audit is more of a warning then a blocker in the pipeline. For this project I imagined an `audit --fix`  being a periodic thing as it may lead to breaking changes. 

Happy to add to production.yml and or strengthen it to a blocker here if y'all feel that would be best. @tonyrosario @timchaffee @kaizengrowth

tonyrosario (dismissed):

package.json
```diff
@@ -57,6 +57,6 @@
     "typescript": "^5"
   },
   "lint-staged": {
-  "**/*.{ts,tsx}": ["eslint --fix"]
+    "**/*.{ts,tsx}": ["eslint --fix"]
```
tonyrosario, line 60: NIT:  needs indent on line 60.  (we moved 59 back ✅  but 60 was moved as well ❌ )

I wasn't going to push it because it doesn't break anything but I was able to verify that the pre-commit hook and the push hook is working! 🙌 

EDIT: You should be able to approve and merge now since I had the last commit

tonyrosario (approved):

timchaffee (approved):
LGTM - please review @tonyrosario comments

PR #70, opened 2026-04-07
fix(dupe-messages): only display supabase msg events

### LINKED ISSUE #85: Optimistic/Realtime dedup race condition in ConversationView.tsx (State bug)

### Steps to reproduce

**Expected:**

**Actual:**
- Image

### Intended fix

xdhmoore:
Instead of manually managing the local messages list and also displaying events once supabase has issued an event (presumably with a new id that doesn't match our local id), this changes it so that we just rely on supabase events. All of a user's new posts to the conversation window will come from supabase events.

I have not been able to reproduce and test this, but I believe @GeorgeDover has, if he can weigh in.

If more testing is needed, I can follow up next weekend when I have more time.

### DIFF
--- a/components/messages/ConversationView.tsx
+++ b/components/messages/ConversationView.tsx
@@ -321,20 +321,6 @@ export function ConversationView({
       messageType = "file";
     }
 
-    const optimisticId = crypto.randomUUID();
-    const optimisticMsg: MessageWithSender = {
-      id: optimisticId,
-      conversation_id: conversationId,
-      sender_id: currentUser.id,
-      content: text,
-      message_type: messageType,
-      metadata,
-      edited_at: null,
-      created_at: new Date().toISOString(),
-      sender: buildSenderProfile(currentUser),
-    };
-    setMessages((prev) => [...prev, optimisticMsg]);
-
     // Use the API route so the server can fire email notifications to other participants
     let savedMsg: Message | null = null;
     let errorMsg: string | null = null;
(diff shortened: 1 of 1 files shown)


GeorgeDover:
I was able to reproduce the error in dev via a breakpoint within handleSend, creating a latency between server response and client's "optimistic" handling of message. Though this is around the wrong way (in prod the client would be first followed by slow server response), the effect is the same - duplicated messages logged in the chat.

I tested this with @xdhmoore's fix above and the problem was resolved. I think it's good to move ahead!

GeorgeDover (approved):

components/messages/ConversationView.tsx
```diff
@@ -321,20 +321,6 @@ export function ConversationView({
...
-      content: text,
-      message_type: messageType,
-      metadata,
-      edited_at: null,
-      created_at: new Date().toISOString(),
-      sender: buildSenderProfile(currentUser),
-    };
-    setMessages((prev) => [...prev, optimisticMsg]);
-
     // Use the API route so the server can fire email notifications to other participants
```
tonyrosario, line 324: fixed confirmed as well.

Do we think that the git commit history is enough for a future contributor to avoid adding back an optimistic insert?  or do we add another comment above 324?

```
// Message display is handled by the Realtime subscription — no optimistic insert needed.
// If restoring optimistic updates, deduplicate against incoming Realtime events to avoid
// the duplicate bug fixed in PR-70.
// Use the API route so the server can fire email notifications to other participants
```

tonyrosario (approved):

components/messages/ConversationView.tsx
xdhmoore: I wouldn't rely on git history, but I think if people are going to call `setMessages()` in this file, they will likely look and see where it is being called in the same file.

PR #71, opened 2026-04-08
Feature/anonymous survey

kaizengrowth:
Thank you @tonyrosario -- we just had the call with everyone, and I'm sending this link out now


app/survey/page.tsx
```diff
@@ -0,0 +1,39 @@
...
+    type: "website",
+  },
+};
+
+export default async function SurveyPage() {
+  // Auth is optional — survey is public. User is passed to nav for logged-in state only.
+  const supabase = await createClient();
+  const {
+    data: { user },
+  } = await supabase.auth.getUser();
```
timchaffee: Should we remove the login graphic or something - I am not 100% sure I would trust it was not linked if it still shows me logged in.

components/survey/survey-trust-section.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <ul className="space-y-2 text-sm text-muted-foreground">
+        <li className="flex gap-2">
+          <span className="mt-0.5 shrink-0 text-green-600">✓</span>
+          <span>No account or login required.</span>
+        </li>
+        <li className="flex gap-2">
+          <span className="mt-0.5 shrink-0 text-green-600">✓</span>
+          <span>
+            Your survey answers (grievances, team, stories) are stored separately from your
+            contact information. They cannot be linked to you even if you provide a contact.
```
timchaffee: I am not sure we should have create_at at all in the DB - If we only got 1 per hour they could be linked just by the submission being the only ones that hour.

components/survey/survey-form.tsx
```diff
@@ -0,0 +1,367 @@
...
+        >
+          Back
+        </Button>
+
+        {step !== "s4" ? (
+          <Button onClick={handleNext}>Next</Button>
+        ) : (
+          <Button
+            onClick={handleSubmit}
+            disabled={isSubmitting || !turnstileToken}
```
timchaffee, line 369: In preview this button was disabled - I am guessing since the turnstile env var is missing. I wonder if we should hide the button and have a config error if its not there. If we could do it earlier it would be better. I know once  its configured and setup this should not happen but I am thinking about 2am troubleshooting.

components/survey/survey-form.tsx
```diff
@@ -0,0 +1,367 @@
+"use client";
+
+import { useState, useEffect, useRef } from "react";
+import { Turnstile, type TurnstileInstance } from "@marsidev/react-turnstile";
```
timchaffee, line 4: Was the code for this reviewed to make sure its safe? Or is there another good way to verify NPM packages are safe?

lib/actions/survey.ts
```diff
@@ -0,0 +1,277 @@
...
+      contactType = data.contactType as ContactType;
+    }
+
+    // Encrypt contact if provided
+    let encryptedContact: string | null = null;
+    let contactIv: string | null = null;
+    let domainHash: string | null = null;
+    const keyVersion = 1;
+
+    if (contactRaw.length > 0 && contactType) {
```
timchaffee, line 393: What is the plan to be able to get unencrypted contact info to the folks that want to contact those who ask to be contacted?

components/survey/survey-form.tsx
tonyrosario: so for testing, [you can use one of their test keys](https://developers.cloudflare.com/turnstile/troubleshooting/testing/)
```
TURNSTILE_SECRET_KEY=1x0000000000000000000000000000000AA
NEXT_PUBLIC_TURNSTILE_SITE_KEY=1x00000000000000000000AA
```

app/survey/page.tsx
tonyrosario: Agree.  Done in a933ccf4c9

Not an impl issue but a visual issue.  People should no longer be logged-in on the survey page.

Updated the text to recommend a private window, that way the session cookie won't be sent with the submission.

components/survey/survey-form.tsx
tonyrosario: No vulnerabilities found from `npm audit` and it seems to be actively maintained.

We could pin it to 1.5.0. to avoid supply-chain attacks.

Can always roll our own if nec.

lib/actions/survey.ts
tonyrosario: Was imagining we could have an admin console.

If we pin down the use case(s), we can discuss options!

components/survey/survey-form.tsx
timchaffee: 👍

components/survey/survey-form.tsx
timchaffee: I think we should PIN it. An attack I have heard about is bad actors buying or taking over a popular project and putting in malicious code.

lib/actions/survey.ts
timchaffee: I am wondering if we should have a seperate command line app that only authorized folks get access vs. having it visible on the web and it somehow getting exposed. But an app could get out in the wild too - but if its an app that still requires access to a key in the env. or supabase then I think it would be safer. But I am not sure the best route. Not something that needs to be solved to get the survey out. And I think we could move the key to vault after we go live as long as we use the same key.

tonyrosario:
Going to close this PR in favor of: https://github.com/What-We-Will/community-platform/pull/72

components/survey/survey-form.tsx
tonyrosario: done ec026b77

GeorgeDover:
Is this PR still needed now that we are reducing some features of the platform? We could revisit the idea in the future with a fresh branch perhaps.

PR #75, opened 2026-04-09
State: Approved members stay on the "Approved" section until manual refresh 

scottbustos:
Fix: Approved members stay on pending screen until manual refresh                                                                                                  
                                                                                                                                                                     
Reduces the polling interval on /pending-approval from 30 seconds to 3 seconds, so approved members are automatically redirected to the dashboard near-instantly after an admin approves them.                                                                                                                                      
                                                                                                                                                                     
  What changed                                                                                                                                                       
  - CHECK_INTERVAL_MS: 30_000 → 3_000 in app/pending-approval/page.tsx                                                                                               
                                                                                                                                                                     
  Testing                                                                                                                                                            
  1. Log in as a pending user => you should land on /pending-approval                                                                                                 
  2. In a separate admin session, approve the user                                                                                                                   
  3. The pending screen should redirect to /dashboard within ~3 seconds                                                                                              

Future consideration:                                                                                                                                               
As membership grows, the polling approach will generate increasing DB load (every active pending user hits the DB every 3 seconds). When traffic scales, this should be migrated to a Supabase Realtime subscription on the user's profile row, the redirect would then be instant with zero wasted queries.

### DIFF
--- a/app/pending-approval/page.tsx
+++ b/app/pending-approval/page.tsx
@@ -6,7 +6,7 @@ import { Button } from "@/components/ui/button";
 import { createClient } from "@/lib/supabase/client";
 import { useRouter } from "next/navigation";
 
-const CHECK_INTERVAL_MS = 30_000; // check every 30 seconds
+const CHECK_INTERVAL_MS = 3_000; // check every 3 seconds
 
 export default function PendingApprovalPage() {
   const router = useRouter();
@@ -34,7 +34,7 @@ export default function PendingApprovalPage() {
     }
   }
 
-  // Poll for approval every 30 seconds
+  // Poll for approval every 3 seconds
   useEffect(() => {
     checkApprovalStatus();
     const interval = setInterval(checkApprovalStatus, CHECK_INTERVAL_MS);
--- a/app/pending-approval/page.tsx
+++ b/app/pending-approval/page.tsx
@@ -6,7 +6,7 @@ import { Button } from "@/components/ui/button";
 import { createClient } from "@/lib/supabase/client";
 import { useRouter } from "next/navigation";
 
-const CHECK_INTERVAL_MS = 3_000; // check every 3 seconds
+const CHECK_INTERVAL_MS = 10_000; // check every 10 seconds
 
 export default function PendingApprovalPage() {
   const router = useRouter();


app/pending-approval/page.tsx
```diff
@@ -6,7 +6,7 @@ import { Button } from "@/components/ui/button";
 import { createClient } from "@/lib/supabase/client";
 import { useRouter } from "next/navigation";
 
-const CHECK_INTERVAL_MS = 30_000; // check every 30 seconds
+const CHECK_INTERVAL_MS = 3_000; // check every 3 seconds
```
timchaffee: That will be alot of DB requests checking if the row is update. I think it would be better to subscribe to that row update in supabase: [https://supabase.com/docs/guides/realtime/subscribing-to-database-changes](https://supabase.com/docs/guides/realtime/subscribing-to-database-changes).  I am not sure how much work it is but we should shy away from polling the database for changes/updates when possible to decrease the load on the DB server.

timchaffee: I see you mentioned that in the description. And I agree - perhaps setting the polling to 10 seconds for now would be okay as 3 seconds could create a lot of traffic if we have a bunch of folks all register and leave that screen up. 10 might be a decent balance until it can be changed to a subscription. Please also add an issue to the Issues to do that work. A pattern I like to use whenever I do something for quick delivery that needs to be optimized later is always add an ISSUE for the future work. That makes it visible and requires it to be prioritized with all the other work.

timchaffee: Once you add an Issue you can mention this PR and OG Issue (if there is one) in the new Issue and put a comment here and mark this as resolved.

scottbustos: Sounds good Tim! I agree that 3 seconds seems a little aggressive. I don't expect the approval workflow to be super heavy at first, but once traffic increases and we're entering a state where the DB is under load, we should migrate this to a Realtime subscription. I've updated the interval to 10 seconds and opened an issue to track the future work: #77

timchaffee: Nice

timchaffee (approved):

PR #82, opened 2026-04-12
fix(email-domain): set NEXT_PUBLIC_SITE_URL correctly in prod

xdhmoore:
I am still testing this, but this should fix the issue where prod emails contain links using the vercel domain. 

There is a chance this could break prod if the config files don't work the way I expect them to.

### DIFF
--- a/.env.preview
+++ b/.env.preview
@@ -1,0 +1,2 @@
+# .env.production
+NEXT_PUBLIC_SITE_URL=
--- a/.env.production
+++ b/.env.production
@@ -1,2 +1,2 @@
 # .env.production
-# NEXT_PUBLIC_SITE_URL is derived in next.config.ts from VERCEL_URL
+NEXT_PUBLIC_SITE_URL=https://wwwrise.org


xdhmoore:
Things to be tested:
[ ] preview app still loads up
[ ] preview emails have the right link
[ ] prod app still loads up
[ ] prod emails have the right link

Note that group msg notification emails are only  sent if the recipient hasn't been online in the last day

xdhmoore:
Tagging @timchaffee on this. Mostly to confirm I'm not breaking prod.

timchaffee:
> Tagging @timchaffee on this. Mostly to confirm I'm not breaking prod.

I am not sure we are even using env.production in production. Right now all the variables seem to be set in the UI. I'll add this one to prod and see if it fixes the email bug.

timchaffee:
This is now set in production: image

timchaffee:
> This is now set in production: image

This appears to have fixed it - I am not sure we need this PR - I don't think it will hurt but we should probably look at if/what we would want in total in a env.production file. It would be very handy to have if we ever needed to deploy to somewhere outside vercel. However, we need to not store any secrets here either.

xdhmoore:
Responding to several things:
1. This PR - I agree we do not need it
2. env.production file - IMHO it is a best practice to set environmental variables (except maybe secrets) in .env files instead of in the vercel UI so that they are under source control and the interplay between the env variable priority is easy to see. It also makes deploys more reproducible. But I would argue that fixing that is out of scope for this task.
3. Testing - we've tested that prod works, but I think we still need to test that preview environments work and maybe dev environments. I will work a little on that. But that can be captured in the Asana task, not here.

PR #93, opened 2026-04-24
fix: move router.refresh() outside state updater in ConversationList

### LINKED ISSUE #85: Optimistic/Realtime dedup race condition in ConversationView.tsx (State bug)

tonyrosario:
Calling router.refresh() inside the setConversations updater triggers a
  Router state update during React's render cycle, producing:
  'Cannot update a component (Router) while rendering ConversationList'.

  This happens when a Realtime INSERT arrives for a conversation not in the
  list (idx === -1). Move the refresh call after the updater completes
  using a flag variable.

  Refs #85
  Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/components/messages/ConversationList.tsx
+++ b/components/messages/ConversationList.tsx
@@ -90,15 +90,16 @@ export function ConversationList({
         { event: "INSERT", schema: "public", table: "messages" },
         (payload) => {
           const newMsg = payload.new as Message;
+          let needsRefresh = false;
 
           setConversations((prev) => {
             const idx = prev.findIndex(
               (c) => c.conversation.id === newMsg.conversation_id
             );
 
-            // Unknown conversation — refresh from server as a fallback.
+            // Unknown conversation — flag for refresh after state update.
             if (idx === -1) {
-              router.refresh();
+              needsRefresh = true;
               return prev;
             }
 
@@ -114,6 +115,10 @@ export function ConversationList({
             updated.splice(idx, 1);
             return [conv, ...updated];
           });
+
+          if (needsRefresh) {
+            router.refresh();
+          }
         }
       )
       .subscribe();
--- a/components/messages/ConversationList.tsx
+++ b/components/messages/ConversationList.tsx
@@ -90,18 +90,21 @@ export function ConversationList({
         { event: "INSERT", schema: "public", table: "messages" },
         (payload) => {
           const newMsg = payload.new as Message;
-          let needsRefresh = false;
+
+          // Unknown conversation — refresh from server to fetch its metadata.
+          const isKnownConversation = conversations.some(
+            (c) => c.conversation.id === newMsg.conversation_id
+          );
+          if (!isKnownConversation) {
+            router.refresh();
+            return;
+          }
 
           setConversations((prev) => {
             const idx = prev.findIndex(
               (c) => c.conversation.id === newMsg.conversation_id
             );
-
-            // Unknown conversation — flag for refresh after state update.
-            if (idx === -1) {
-              needsRefresh = true;
-              return prev;
-            }
+            if (idx === -1) return prev;
 
             const updated = [...prev];
             const conv = { ...updated[idx] };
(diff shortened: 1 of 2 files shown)


components/messages/ConversationList.tsx
```diff
@@ -114,6 +115,10 @@ export function ConversationList({
             updated.splice(idx, 1);
             return [conv, ...updated];
           });
+
+          if (needsRefresh) {
```
xdhmoore: The way this is written currently it will first run line 119-121, and then later get around to running the function passed into `setConversations` and setting `needsRefresh`. I'm new to React, but I think you want something like this instead of lines 119-121:

```
useEffect(() -> {
   router.refresh();
}, [conversations]);
```
But this feels wrong to me too. Shouldn't the component already be reloading when `conversations` changes? Why do we need to force a router reload?

tonyrosario: New to react as well :)

Agree that we want to confirm the execution order.

From my research, the updater function to `setConversations` is synchronous, React calls it inline before it
returns. So by the time we hit the if (needsRefresh) check, the flag has already been set by the updater. The
re-render is batched for later, but the updater itself executes immediately. The flag sits right between those
two moments, which is where we need it.

On the `useEffect` idea,  the instinct makes sense, but it doesn't fit here because when the conversation is
unknown (idx === -1), we return prev unchanged. No state change = no re-render = the effect wouldn't fire. And if we made it fire on all conversation changes, it would trigger router.refresh() on every normal message update too, which we don't want.

To the question "shouldn't the component already be reloading when conversations changes?" — this is a key React distinction: re-rendering ≠ reloading data. When state changes, React re-runs the component function and updates the DOM, but only with data already in memory. It doesn't go back to the server. In the
unknown-conversation case, we have a message but not the conversation metadata (participants, display names, etc.) needed to render a new list item — that data lives on the server. router.refresh() is what tells Next.js to re-run the Server Component parent, re-fetch from Supabase, and send the full conversation list back down. For known conversations we skip this entirely since we already have everything we need client-side.

> Disclosure: I leaned on Claude Code to help me validate the React execution model and draft parts of this explanation (mainly last 2 paragraphs) — per our AI policy.

tonyrosario: I'm taking a closer look based on `The flag sits right between those two moments, which is where we need it.`

During this fix, I conceded that some tests would be covered by an E2E Playright test but making this comment didn't sit well with me - **needing** a bit of code in a specific place/position screams fragility 😱.

I'm going to refactor this real quick and push up a change.

xdhmoore (dismissed):
Head's up on the react hook warning on line 128, but looks good otherwise.

components/messages/ConversationList.tsx
xdhmoore: Despite getting conflicting results from chatgpt, I think this line in the react docs is helpful about the asynchronous execution of setters:

> React puts your updater functions in a [queue.](https://react.dev/learn/queueing-a-series-of-state-updates) Then, during the next render, it will call them in the same order:

It sounds to me like they are run asynchronously. But it looks like your next change sidelines the issue.

from: https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

tonyrosario: @xdhmoore You're right, my previous explanation was wrong. Checking the docs confirmed it: updater functions are queued and run during the next render, not synchronously inline... which is why the flag pattern didn't work.

The latest change attempts to address the timing problem: read current state from a ref, then decide whether to update, before entering the updater.

Thanks for spotting this!

> Disclosure: leaned on Claude Code to validate the React execution model

tonyrosario:
> Head's up on the react hook warning on line 128, but looks good otherwise.

Should be fixed now.  Ready for re-review

SirFuzzyWalls (approved):
👍

PR #94, opened 2026-04-24
test: add regression coverage for ConversationView Realtime INSERT ha…

### LINKED ISSUE #85: Optimistic/Realtime dedup race condition in ConversationView.tsx (State bug)

tonyrosario:
…ndler

PR #70 removed the optimistic prepend from handleSend — the sender's own message now enters state only via the Realtime subscription. Add unit tests that lock down two behaviors: no optimistic path on send, and deduplication when the same INSERT arrives twice.

Refs #70, #85

### DIFF
--- a/components/messages/__tests__/ConversationView.test.tsx
+++ b/components/messages/__tests__/ConversationView.test.tsx
@@ -1,0 +1,186 @@
+import { render, screen, act, fireEvent } from "@testing-library/react";
+import { ConversationView } from "../ConversationView";
+import type { Profile } from "@/lib/types";
+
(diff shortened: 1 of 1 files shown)


xdhmoore (dismissed):
LGTM

tonyrosario:
@xdhmoore I pushed an update to the tests after finding an issue in #93.  Updated some test descriptions while I was there too

xdhmoore (approved):

xdhmoore (approved):

PR #95, opened 2026-04-27
Feat/76 infra add e2e testing

### LINKED ISSUE #76: Infra: Add End-to-End testing with Playwright
**Context**: End-to-end testing is currently a manual process, that sometimes requires singing up with a new account and navigating around the platform to confirm no breaking changes were made.

**Solution**: Add Playwright testing to well worn user paths, setting a standard for how future tests are written.

tonyrosario:


playwright.config.ts
```diff
@@ -0,0 +1,46 @@
+import { defineConfig } from "@playwright/test";
+import { existsSync } from "node:fs";
+
+// Load E2E test credentials from .env.e2e (never committed).
+// Uses Node 20.12+ built-in — no dotenv dependency needed.
+if (existsSync(".env.e2e")) {
+  process.loadEnvFile(".env.e2e");
+}
```
tonyrosario, line 8: needs node 20+ for this to work.

Are we all using that?  because there isn't a pin in package.json

kaizengrowth:
Awesome work on this!

e2e/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+Fill in the six values with the emails and passwords from step 2. `.env.e2e` is gitignored — never commit it.
+
+### 5. Run the tests
+
+```bash
+npm run test:e2e
+```
+
+Expect **12 passed**. If the auth specs skip, your `PW_E2E_*` env vars aren't being picked up — check that `.env.e2e` is in the repo root.
```
timchaffee: I am not sure we should list 12 tests since if we add more this number will change - would it be possible to get this to run just the auth tests?

TESTING_STANDARDS.preamble.md
timchaffee, line 1: Should we add anything to this file about requiring Playwright tests?

timchaffee (dismissed):
LGTM - with comments

e2e/README.md
tonyrosario: Good catch - I thought I fixed that but either I missed it or this is another instance.

I'll reach out on slack for the second bit

e2e/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+Both `test-results/` and `playwright-report/` are gitignored.
+
+## First-time setup
+
+The auth specs need real users in **your own Supabase project**. Never point these at the shared preview or production environments.
+
+### 1. Point your local app at your own Supabase project
+
+Set `NEXT_PUBLIC_SUPABASE_URL` and `NEXT_PUBLIC_SUPABASE_ANON_KEY` in `.env.local` to your personal project. If you don't have one, create a free project at [supabase.com](https://supabase.com), run the migrations in `supabase/migrations/` against it, and use those credentials.
```
SirFuzzyWalls, line 97: NIT: Let's include a bash command here, or a link to the readme. 

```bash
supabase link --project-ref <your-project-ref>
supabase db push
```

TESTING_STANDARDS.preamble.md
SirFuzzyWalls: Not sure on this, my concern is that we wind up reaching for playwright tests on everything. Perhaps a middle ground - if your branch touches auth, login flow, or other critical paths?

Then again, playwright tests on everything sounds like a utopia. I'll defer to the wisdom of folks who've lived with playwright before. CC @scottbustos @GeorgeDover

SirFuzzyWalls (dismissed):
Approved with very small comments around docs. Great stuff and so freaking cool!

e2e/README.md
tonyrosario: next commit has the link to readme!

tonyrosario: @SirFuzzyWalls I just pushed the updated readme in b0440f3359

e2e/README.md
tonyrosario: Should be fixed in the latest commit b0440f3359

TESTING_STANDARDS.preamble.md
tonyrosario: My .02 is that it shouldn't be required because 1. not every PR will affect front-end functionality and 2. they are costly to maintain and can be very brittle.

We should focus on having more unit tests in the code base and keep the e2e suite small by comparison (a-la test pyramid).

However, I do think coverage on the most critical points of the site would be beneficial, like getting people approved, onboarded, and interacting with the platform would be good examples for this project.

We'll take that list and balance it with parts of the code-base that are NOT being actively updated. This will help us avoid the cycle of constantly fixing e2e tests.

Suggest that we create another issue and then plan out what we want covered in the short-term, critical paths, anonymous-access checks, and RLS-at-the-UI-layer.

EDIT: I think we can add the guidance on what needs these e2e tests as part of the acceptance criteria in the next issue.

timchaffee: In the end I have seen great value in E2E tests for everything UI - doing that gives you a very high confidence you can deploy anytime without introducing new or regression bugs.  And allows for CD - every merge to main deploys as long as unit and E2E test pass. We are not there yet and I would concnetrate on unit tests and playright tests for new functionality. And as we have time add more and more PW tests to cover the UI.

timchaffee (approved):
LGTM

TESTING_STANDARDS.preamble.md
tonyrosario: Agreed.  We can always adjust based on our needs.  I'll merge this for now and create the next issue so that we can see how this plays out.  We can always adjust our strategy once we start to get some feedback from writing + using the tests.

PR #98, opened 2026-04-30
Changes to MemberFilters Component

### LINKED ISSUE #68: Bug: Small Bugs - Low Priority
Here is a list of current bugs I have seen and will plan to fix. I will add anymore I find. This is low priority. Please add any small user issues here as well. MAY HIGH PRIORITY, searching for members takes up a lot of request on the supabase backend costing resources.

- [x] Messages double send
- [x] The member search bugs out when backspacing

Savern02:
fixes bug found in open issue https://github.com/What-We-Will/community-platform/issues/68

### DIFF
--- a/components/members/MemberFilters.tsx
+++ b/components/members/MemberFilters.tsx
@@ -15,26 +15,27 @@ import {
 
 const SEARCH_DEBOUNCE_MS = 300;
 
-export default function MemberFilters({ allSkills }) {
+type MemberFiltersProps = {
+  allSkills: string[];
+};
+
+export default function MemberFilters({ allSkills }: MemberFiltersProps) {
   const router = useRouter();
   const searchParams = useSearchParams();
 
-  // make sure there is only one local state 
-  const [q, setQ] = useState(() => searchParams.get("q") ?? "");
-  const [skill, setSkill] = useState(() => searchParams.get("skill") ?? "");
-  const [referrals, setReferrals] = useState(
+  const [q, setQ] = useState<string>(() => searchParams.get("q") ?? "");
+  const [skill, setSkill] = useState<string>(() => searchParams.get("skill") ?? "");
+  const [referrals, setReferrals] = useState<boolean>(
     () => searchParams.get("referrals") === "true"
   );
 
-  //debounce only ONE value (q)
-  const [debouncedQ, setDebouncedQ] = useState(q);
+  const [debouncedQ, setDebouncedQ] = useState<string>(q);
 
   useEffect(() => {
     const t = setTimeout(() => setDebouncedQ(q), SEARCH_DEBOUNCE_MS);
     return () => clearTimeout(t);
   }, [q]);
 
-  //sync URL (single controlled effect)
   useEffect(() => {
     const params = new URLSearchParams();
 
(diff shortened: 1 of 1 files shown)


components/members/MemberFilters.tsx
```diff
@@ -15,112 +15,75 @@ import {
...
+    router.replace(url);
+  }, [debouncedQ, skill, referrals, router]);
 
   return (
     <div className="flex flex-col gap-4 sm:flex-row sm:items-end">
       <div className="flex-1 space-y-2">
-        <Label htmlFor="search">Search</Label>
+        <Label>Search</Label>
         <Input
-          id="search"
```
SirFuzzyWalls, line 84: If possible, I'd like to keep the accessibility  attributes such as `id="search"` `type="search"`. I believe they have knock on impacts for mobile page views and screen readers.

SirFuzzyWalls:
First, great stuff! I've been using the preview env for this and it's working well! Congrats!


**Love to see:** Unit tests for this using existing styles and components (eg `login-form.test.tsx`)

We've had several issues with this particular functionality  #108 is a dupe of this. 

I will dig a little further on this in the evening.

components/members/MemberFilters.tsx
```diff
@@ -15,112 +15,75 @@ import {
...
         />
       </div>
+
       <div className="space-y-2">
         <Label>Skill</Label>
-        <Select
-          value={skill || "all"}
-          onValueChange={(v) => updateParams({ skill: v === "all" ? "" : v })}
-        >
+        <Select value={skill || "all"} onValueChange={setSkill}>
```
tonyrosario, line 66: Removed based on [my comment ](https://github.com/What-We-Will/community-platform/pull/98#issuecomment-4360304338)

~~skill=all leaks to URL which leads to empty results.~~

~~Fix: `onValueChange={(v) => setSkill(v === "all" ? "" : v)}`~~

~~Steps to repro (before fix):~~
  1. ~~go to /members~~
  2. ~~Pick any specific skill from the dropdown — note results filter~~
  3. ~~Pick "All skills"~~
 4.  ~~Bug: URL becomes /members?skill=all, member list goes empty (with "No members found")~~
 5.  ~~After fix: URL becomes /members (no skill param), full list returns~~

~~NOTE: Could be a candidate for a e2e test.  Let me know if you'd like to try to add it or I can push a spec.~~

tonyrosario:
After reviewing, this PR deviates from an established pattern for filter components found in EventFilters and JobRoleFilter:
1. URL as source of truth
2. Single updateParams helper
3. router.push

The pre-PR version of MemberFilters actually followed this pattern too. This refactor moves away from all three (local state, router.replace, inline URL building), and that divergence is what's causing the back/forward and skill=all bugs.
 
Can we align with the sibling pattern instead? Happy to keep local state for q since the debounce justifies it — no other filter has a free-text input.

Happy to put together a test harness on a separate branch if it'd help. Let me know!

components/members/MemberFilters.tsx
```diff
@@ -15,112 +15,75 @@ import {
...
-  // Sync URL when debounced search changes
-  useEffect(() => {
-    const currentQ = searchParams.get("q") ?? "";
-    if (debouncedSearch !== currentQ) {
-      updateParams({ q: debouncedSearch });
-    }
-  }, [debouncedSearch, updateParams, searchParams]);
+    const url = `/members${params.toString() ? `?${params.toString()}` : ""}`;
+
+    router.replace(url);
```
tonyrosario, line 48: yes, keep this but please note in the PR description why

components/members/MemberFilters.tsx
tonyrosario: Agree.  I see 3.

lines 54, 65, 86 in the newly modified source.

tonyrosario (changes requested):

Issue #78, opened 2026-04-10
Dev Experience: Resolve concurrent migration conflicts


tonyrosario:
## Problem
When multiple developers create migrations concurrently, sequential migration numbers conflict (e.g., two PRs both generate `0042_...`), causing CI failures and broken migration histories on merge.

This is actively affecting contributors working in parallel.

## Current Constraints
- 2 databases available on current plan (dev + prod)
- Upgrading the plan is an option if needed

## Questions to Resolve
- What naming/process strategy should we adopt to prevent conflicts?
- Do we need additional databases, or can we solve this with process/tooling alone?

## Acceptance Criteria
- [x] Two developers can open concurrent PRs with migrations without conflicts
- [x] Solution is documented in ~~CONTRIBUTING.md~~ README.md

tonyrosario:
Super-quick research - still need to review and ask around what other options are available.

## Option 1: Timestamp-based migrations (easiest, free)
Switch from sequential (`0001`, `0002`) to timestamp-based names (`20240410143200_add_users.sql`). Supabase CLI supports this natively.
```
supabase migration new add_users_table
# generates: 20240410143200_add_users_table.sql
```
**Why it helps:** Collisions are near-impossible since timestamps are to the second. Ordering is still deterministic. This is the lowest-effort fix and works within your current setup.

**Caveat:** If two devs create a migration at the exact same second (rare), you still conflict — but this is extremely uncommon.


## Option 2: Branch-per-developer workflow (free tier workaround)
Since you only have 2 databases, use one as production and one as a shared dev/staging. Developers:

Work locally with `supabase start` (local Docker Postgres — fully free, unlimited)
Only push migrations to the shared dev DB when merging to a shared branch
Never write migrations against a shared DB directly

**The rule:** Migrations are written locally, reviewed in PRs, and applied in merge order. This is the discipline fix — no tooling required.

## Option 3: Supabase branching (requires paid plan)
Supabase has a **Branching** feature (Pro plan, ~$25/month) that gives each PR/branch its own ephemeral database instance. This is the "proper" solution:

Each PR gets an isolated DB branch
Migrations are tested in isolation
No stepping on each other at all

This is worth it at 5-7 developers. The cost is low and the DX improvement is significant.

## Option 4: Migration serialization via CI (free, more process)
Enforce a rule in CI: migrations can only be merged to `main` one at a time, and the branch must be rebased/up-to-date before merge. You add a CI check like:
```
[...]

PR #43, opened 2026-03-25
Fixes #41 Chore/testing standards

tonyrosario:
Initial document for Testing Standards for review.

However before it is final, we'll need to align on some questions in the [TESTING_DECISIONS.md](https://github.com/What-We-Will/community-platform/blob/ddbcf4603173c130441e504c235fccb0d66af9ac/TESTING_DECISIONS.md)

PR #127, opened 2026-05-06
Feat/intro to adr

tonyrosario:
We're starting to make a decent amount of decisions.  Thought ADR's would be a good way to capture these decisions as we go along.

I don't expect every change/PR to need an ADR but rather for 'big' changes like #78 or #43.  Don't want to lose the context over time.

### DIFF
--- a/docs/adr/README.md
+++ b/docs/adr/README.md
@@ -2,7 +2,7 @@
 
 This directory holds the project's Architecture Decision Records — short, durable notes capturing decisions that are load-bearing, hard to reverse, or already came out of substantive debate.
 
-ADRs are based on [Michael Nygard's format](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions): Title, Status, Context, Decision, and Consequences. This project adds three header fields (`TL;DR`, `Author`, and `Sponsoring Lead`), a required `Alternatives considered` body section, an optional `Open questions` body section, and one drafting status (`Draft`).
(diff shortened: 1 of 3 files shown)


docs/adr/README.md
```diff
@@ -0,0 +1,136 @@
...
+| Lead | GitHub | Primary ownership area |
+|---|---|---|
+| Simon McGraw | @SirFuzzyWalls | infra, data model, RLS, migrations |
+| Tim Chaffee | @timchaffee | frontend, CI/CD, accessibility |
+| Tony Rosario | @tonyrosario | design system, observability |
+
+Ownership areas are guidance for routing, not hard ownership. Any lead can sponsor any ADR. The table exists so a volunteer at 11pm on a Saturday knows who to @-mention without having to ask in chat first.
+
+> [!IMPORTANT]
+> **Provisional rule — resolve before merge (2026-05-06).** The 5-business-day window below has not been confirmed by all three leads. @SirFuzzyWalls @timchaffee — please confirm or propose a different value in this PR. Remove this callout once consensus lands.
```
tonyrosario: This is one outstanding issue.  Trying to find a balance between an issue from becoming stale but not introducing a process that will slow us down.

Forgive the wording, it doesn't need to be "confirmed by all three leads" but was looking for something to say that someone reviewed it.

See also below, line 85.  Is PR thread enough? Should they ping in Slack to escalate?

tonyrosario: Going to merge as-is because I need to start capturing some decisions.  Happy to make any modifications or to review any suggestions in another PR!

supabase/migrations/CONVENTIONS.md
```diff
@@ -0,0 +1,57 @@
+# Migration Conventions
+
+Durable rules for `supabase/migrations/`. Plans live in `plans/local/`; the rules in this file outlive any single plan.
```
tonyrosario: Looks like I need to remove this...  note about supabase/migrations.  

Relates to Issue #78

tonyrosario:
failing until https://github.com/What-We-Will/community-platform/pull/134 gets merged so that it skips the supabase tasks. See Issue #133

PR #129, opened 2026-05-07
Resolve merge conflict and update messages + migration + deletion fix

OmegaDeanGrey:
Message delete

In ConversationsList

Line 13:
Added the import: Trash2

Line 124:
const [isDeleting, setIsDeleting] = useState<string | null>(null);

Lines 126-156:
async function handleDelete(
  e: React.MouseEvent,
  conversationId: string
) {
  e.preventDefault();
  e.stopPropagation();

  setIsDeleting(conversationId);

  const supabase = createClient();

  await supabase
    .from("conversation_participants")
    .update({
      deleted_at: new Date().toISOString(),
    })
    .eq("conversation_id", conversationId)
    .eq("user_id", currentUserId);

  setConversations((prev) =>
    prev.filter((c) => c.conversation.id !== conversationId)
  );

  if (pathname === `/messages/${conversationId}`) {
    router.push("/messages");
  }

  router.refresh();

  setIsDeleting(null);
}

Line 321-335:
<TooltipProvider>
  <Tooltip>
    <TooltipTrigger asChild>
      <Button
        variant="ghost"
        size="icon"
        className="size-8 shrink-0"
        disabled={isDeleting === conversation.id}
        onClick={(e) => handleDelete(e, conversation.id)}
      >
        <Trash2 className="size-4" />
      </Button>
    </TooltipTrigger>
    <TooltipContent>Delete conversation</TooltipContent>
  </Tooltip>
</TooltipProvider>

Backend:

Created new migration at supabase/migrations/20260501152108_create_message_deletions.sql
 CREATE TABLE IF NOT EXISTS message_deletions (
  id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
  message_id uuid NOT NULL REFERENCES messages(id) ON DELETE CASCADE,
  user_id uuid NOT NULL REFERENCES profiles(id) ON DELETE CASCADE,
  deleted_at timestamptz NOT NULL DEFAULT now(),

  UNIQUE(message_id, user_id)
);

In: community-platform/app/(app)/messages/[conversationId]/page.tsx 
Lines 70-77 const { data: deletions } = await supabase
.from("message_deletions")
.select("message_id")
.eq("user_id", user.id);
  const deletedMessageIds = new Set(
        (deletions ?? []).map((d) => d.message_id)
      );

Lines 118 & 194
Reversed the order of .reverse & .filter 

This is not exact, because of a long merge conflict battle.  Trying to push the cleanest message soft delete I can with no extraneous code.  
- Resolved merge conflict in ConversationList.tsx
- Updated messages page logic
- Added migration for message deletions
- Improved real-time conversation updates

### DIFF
--- a/app/(app)/messages/[conversationId]/page.tsx
+++ b/app/(app)/messages/[conversationId]/page.tsx
@@ -15,7 +15,10 @@ export default async function ConversationPage({
 }: ConversationPageProps) {
   const { conversationId } = await params;
   const resolvedSearchParams = await searchParams;
-  const videoRoom = typeof resolvedSearchParams?.videoRoom === "string" ? resolvedSearchParams.videoRoom : null;
+  const videoRoom =
+    typeof resolvedSearchParams?.videoRoom === "string"
+      ? resolvedSearchParams.videoRoom
+      : null;
 
   const supabase = await createClient();
   const {
@@ -32,7 +35,10 @@ export default async function ConversationPage({
     .maybeSingle();
 
   if (myParticipationError) {
-    console.error("[conversation] myParticipation error:", myParticipationError);
+    console.error(
+      "[conversation] myParticipation error:",
+      myParticipationError,
+    );
   }
 
   if (!myParticipation) {
@@ -40,7 +46,12 @@ export default async function ConversationPage({
     if (conversationId === selfNotesConversationId(user.id)) {
       await getOrCreateSelfNotes(user.id);
     } else {
-      console.error("[conversation] user", user.id, "is not a participant in", conversationId);
+      console.error(
+        "[conversation] user",
+        user.id,
+        "is not a participant in",
+        conversationId,
+      );
       redirect("/messages");
     }
   }
@@ -66,6 +77,13 @@ export default async function ConversationPage({
     display_name: currentUserProfile?.display_name ?? "You",
     avatar_url: currentUserProfile?.avatar_url ?? null,
   };
+
+  const { data: deletions } = await supabase
+    .from("message_deletions")
+    .select("message_id")
+    .eq("user_id", user.id);
+
+  const deletedMessageIds = new Set((deletions ?? []).map((d) => d.message_id));
 
   // ── Group conversation ────────────────────────────────────────────────────
   if (conversation.type === "group") {
@@ -93,25 +111,37 @@ export default async function ConversationPage({
 
     // Build profile cache for messages
     const profileCache = new Map<string, Profile>(
-      participants.map((p) => [p.id, p])
+      participants.map((p) => [p.id, p]),
     );
 
     const { data: rawMessages } = await supabase
       .from("messages")
       .select(
-        "id, conversation_id, sender_id, content, message_type, metadata, edited_at, created_at"
+        `
+    id, 
+    conversation_id, 
+    sender_id, 
+    content, 
+    message_type, 
+    metadata, 
+    edited_at, 
+    created_at
+  `,
       )
       .eq("conversation_id", conversationId)
       .order("created_at", { ascending: false })
       .limit(50);
 
     const messages: MessageWithSender[] = (rawMessages ?? [])
+      .filter((msg) => !deletedMessageIds.has(msg.id))
       .reverse()
       .map((msg) => ({
         ...msg,
         message_type: msg.message_type as MessageWithSender["message_type"],
         metadata: (msg.metadata ?? {}) as Record<string, unknown>,
-        sender: msg.sender_id ? (profileCache.get(msg.sender_id) ?? null) : null,
+        sender: msg.sender_id
+          ? (profileCache.get(msg.sender_id) ?? null)
+          : null,
       }));
 
     await supabase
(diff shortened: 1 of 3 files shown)


SirFuzzyWalls (changes requested):
After taking a closer look and running this through a review agent, there's two main issues that need to be tackled before we can merge this in.

  **Critical**

  - No [RLS ](https://supabase.com/docs/guides/database/postgres/row-level-security)(Row Level Security)  on message_deletions. Every other table has ENABLE ROW LEVEL SECURITY. Without it, any authenticated user can read or write deletion records for any   other user. Easy fix: Add to the migration. 

```  
ALTER TABLE message_deletions ENABLE ROW LEVEL SECURITY;

  CREATE POLICY "users manage own deletions"
    ON message_deletions
    FOR ALL
    USING (user_id = auth.uid())
    WITH CHECK (user_id = auth.uid());
```


  - No error handling in handleDelete (ConversationList.tsx:126–156) — The supabase .update() result is completely ignored. If the DB call fails, the conversation is optimistically removed from the UI with no feedback and no rollback. Proposed fix:
```
  async function handleDelete(e: React.MouseEvent, conversationId: string) {
    e.preventDefault();
    e.stopPropagation();
    setIsDeleting(conversationId);

    const supabase = createClient();
    const { error } = await supabase
      .from("conversation_participants")
      .update({ deleted_at: new Date().toISOString() })
      .eq("conversation_id", conversationId)
      .eq("user_id", currentUserId);

    setIsDeleting(null);

    if (error) {
      console.error("[conversation] delete failed:", error);
      // surface an error toast here
      return;
    }

    setConversations((prev) =>
      prev.filter((c) => c.conversation.id !== conversationId),
    );

    if (pathname === `/messages/${conversationId}`) {
      router.push("/messages");
    }
  }
```
  - Note: router.refresh() is removed, see the [race condition](https://www.geeksforgeeks.org/operating-systems/race-condition-in-operating-systems/) issue below.
  - Incomplete feature: message_deletions table and the filtering logic in page.tsx:78–84 are wired up with no UI path to populate the table as far as I can see, I need to take another look but I believe this needs a UI element before it's user facing.  

  **Bugs / Nitpicks**

  - router.refresh() races the optimistic update (ConversationList.tsx:148–150). The conversation is removed from local state, then router.refresh() is immediately called. The server re-render re-passes initialConversations to the client, which can  briefly re-add the deleted conversation before the DB write has finished. handleArchive has the same pattern. One option is to remove the router.refresh() calls, the setConversations filter may besufficient, and the real-time subscription will handle any subsequent sync.
  - Delete only added to group rows, not DM rows. The Trash2 button is rendered in the group conversation branch but the DM row only has the Archive button. If deletion should work for both, add the same button block to the DM return:
```
  <TooltipProvider>
    <Tooltip>
          size="icon"
          className="size-8 shrink-0"
          disabled={isDeleting === conversation.id}
          onClick={(e) => handleDelete(e, conversation.id)}
        >
          <Trash2 className="size-4" />
        </Button>
      </TooltipTrigger>
      <TooltipContent>Delete conversation</TooltipContent>
    </Tooltip>
  </TooltipProvider>
```
  - groupSlug destructured but unused, consider removing it from the .map() destructuring to avoid confusion.

  **Migrations**

  - Missing newline at end of file: 20260501152108_create_message_deletions.sql has no trailing newline, may cause a migration issue. 
  
  - deletedMessageIds query is unbounded (page.tsx:78–84) — Fetches all deleted message IDs for the user with no limit. That's fine at the moment but it will grow unbounded. It could be scoped the the current conversation: 
```
  const { data: deletions } = await supabase
    .from("message_deletions")
    .select("message_id")
    .eq("user_id", user.id)
    .in("message_id", (rawMessages ?? []).map((m) => m.id));
```

  **Minor**

  - Silent error: const { data: deletions } = await supabase... discards the error field. If the query fails, deletions is null, the set is empty, and deleted messages silently reappear:
```
  const { data: deletions, error: deletionsError } = await supabase...
  if (deletionsError) console.error("[conversation] deletions fetch error:", deletionsError);
```

  - Multiple TooltipProvider wrappers — Each button in the group row has its own <TooltipProvider>. You can wrap both buttons in a single provider instead:
```
  <TooltipProvider>
    <Tooltip>...</Tooltip> {/* Delete */}
    <Tooltip>...</Tooltip> {/* Archive */}
  </TooltipProvider>
```
  ---
 **The good!** 

  - The .filter() before .reverse() fix looks good! Filtering first avoids reversing items that will be discarded.
  - The message_deletions schema is clean: UNIQUE(message_id, user_id) prevents duplicate records and ON DELETE CASCADE keeps it tidy.
  - isDeleting state correctly disables the button during the async operation.

tonyrosario:
@OmegaDeanGrey quick questions:
1. what was the purpose of the PR?
2. was there an issue related to it?

I think if we lock down the purpose of the PR, then the review can be focused around the feature.  I'm getting a hint that there might be two features in this PR which is why I'm asking just to be clear.

OmegaDeanGrey:
The purpose was to get a soft delete for messages as there currently isn’t a delete.

PR #131, opened 2026-05-12
fix: resolve #116 — Avatar upload size limit too small at 2 MB

### LINKED ISSUE #116: Avatar upload size limit too small at 2 MB
## Description
Profile avatar uploads are limited to 2 MB, which is too small for many modern phone photos.

## How to Reproduce
1. Navigate to your profile
2. Try to upload an image larger than 2 MB as your avatar
3. Upload is rejected

## Expected Behavior
The size limit should be increased (e.g., to 5 or 10 MB) to accommodate typical photo sizes. Client-side image compression before upload would also help.

## Relevant Files
- `components/profile/AvatarUpload.tsx` — `AVATAR_LIMIT_MB = 2`

## Fix
Increase the `AVATAR_LIMIT_MB` constant. Optionally add client-side image resizing/compression before upload.

SlncTrZ:
## Summary

fix: resolve #116 — Avatar upload size limit too small at 2 MB

## Problem

**Severity**: `Medium` | **File**: `components/profile/AvatarUpload.tsx`

The current 2 MB limit is too restrictive for many modern phone photos. This change increases the client-side limit to 10 MB and introduces image compression before upload, reducing the actual file size sent to the server. Compression uses the browser’s Canvas API to resize the image (max 1024x1024) and adjust JPEG quality to ~70%, typically resulting in a file under 1 MB without noticeable quality loss for avatars.

## Changes

- `components/profile/AvatarUpload.tsx` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced"

### DIFF
--- a/components/profile/AvatarUpload.tsx
+++ b/components/profile/AvatarUpload.tsx
@@ -6,8 +6,52 @@ import { cn } from "@/lib/utils";
 import { getAvatarColor, getInitials } from "@/lib/utils/avatar";
 import { uploadPublicFile, validateFile } from "@/lib/storage";
 
-const AVATAR_LIMIT_MB = 2;
+const AVATAR_LIMIT_MB = 10;
 const AVATAR_TYPES = ["image/png", "image/jpeg", "image/webp", "image/gif"];
+
+async function compressImage(file: File): Promise<File> {
+  return new Promise((resolve, reject) => {
+    const img = new Image();
+    img.onload = () => {
+      URL.revokeObjectURL(img.src);
+      let width = img.width;
+      let height = img.height;
+      const maxSize = 1024;
+      if (width > maxSize || height > maxSize) {
+        if (width > height) {
+          height = (height / width) * maxSize;
+          width = maxSize;
+        } else {
+          width = (width / height) * maxSize;
+          height = maxSize;
+        }
+      }
+      const canvas = document.createElement("canvas");
+      canvas.width = width;
+      canvas.height = height;
+      const ctx = canvas.getContext("2d");
+      ctx?.drawImage(img, 0, 0, width, height);
+      canvas.toBlob(
+        (blob) => {
+          if (blob) {
+            const newFileName = file.name.replace(/\.[^/.]+$/, ".jpg");
+            const compressedFile = new File([blob], newFileName, { type: "image/jpeg" });
+            resolve(compressedFile);
+          } else {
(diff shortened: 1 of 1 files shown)


xdhmoore:
That's a neat trick using Canvas. Can we also restrict the size on the server side? Any client side calls can be bypassed for an attack.

I heard a story once about a SharePoint instance that was having issues and it turned out a guy was backing up his whole laptop to a file and uploading it.

components/profile/AvatarUpload.tsx
```diff
@@ -6,9 +6,53 @@ import { cn } from "@/lib/utils";
...
+      }
+      const canvas = document.createElement("canvas");
+      canvas.width = width;
+      canvas.height = height;
+      const ctx = canvas.getContext("2d");
+      ctx?.drawImage(img, 0, 0, width, height);
+      canvas.toBlob(
+        (blob) => {
+          if (blob) {
+            const newFileName = file.name.replace(/\.[^/.]+$/, ".jpg");
```
xdhmoore, line 37: I'm not familiar with how the react file upload element works, but do we need to handle the case where the file name doesn't include a period or ends in a slash? I think currently that will result in a full URL as the file name.

components/profile/AvatarUpload.tsx
```diff
@@ -44,10 +88,18 @@ export function AvatarUpload({
     }
 
     setUploading(true);
-    const ext = file.name.split(".").pop()?.toLowerCase() || "jpg";
+    let uploadFile = file;
+    let ext = file.name.split(".").pop()?.toLowerCase() || "jpg";
```
xdhmoore, line 92: It's using ? here but not for all calls. Would this be better?
```
const ext = file?.name?.split(".")?.pop()?.toLowerCase() || "jpg";
```

components/profile/AvatarUpload.tsx
```diff
@@ -44,10 +88,18 @@ export function AvatarUpload({
...
+    try {
+      const compressed = await compressImage(file);
+      uploadFile = compressed;
+      ext = "jpg";
+    } catch {
+      // Fall back to original file
```
xdhmoore, line 98: Suggestion: do a console.warn() here so we're not leaving an empty catch block. Idk if we're capturing console.warn's but I think it's good practice.

PR #141, opened 2026-05-21
fix: resolve #108 - Member search cycles and jumps between values while typing

### LINKED ISSUE #108: Member search cycles and jumps between values while typing
## Description
When searching in the Members tab, the search field cycles between different versions of the typed input, bouncing back and forth every second. Altering a search also causes visible cycling through adjustments. In some cases, search returns no results even for valid queries.

## How to Reproduce
1. Navigate to the Members tab
2. Start typing in the search box
3. Observe the search text jumping/cycling between values
4. The displayed results flicker as the search bounces

## Expected Behavior
Search input should remain stable while typing and only trigger a search after the user stops typing (debounce).

## Relevant Files
- `components/members/MemberFilters.tsx` (lines 33-77)

## Root Cause
Three-way state sync between `searchInput`, `debouncedSearch`, and URL `searchParams` creates a feedback loop:
1. URL sync effect (line 33-38) resets `searchInput` when `searchParams` changes
2. This triggers the debounce effect (line 44-49) to update `debouncedSearch`
3. Which triggers the URL update effect (line 72-77) to push new params
4. Which triggers step 1 again — creating a loop

## Fix
Make the URL the single source of truth and break the circular dependency between the three state variables.

GeorgeDover:
Closes https://github.com/What-We-Will/community-platform/issues/108. 

## Feedback loop resolution
Search input fields are ordinarily uncontrolled inputs unless there is any drop-down selection UI that populates while typing and can be selected to override the input field.  As the input field in this case is much simpler, we can afford to make it uncontrolled - we remove state management as this causes re-renders when changes occur, and replace with a ref to the input to mutate the DOM directly without React re-rendering.

## `useCallback` > `useEffect`
I also moved the debounce logic inside a `useCallback` event handler which is the pattern React docs recommend over effects. Both work but the former is less prone to unintentional calls due to the explicit attachment of the handler to the component.

## Tests!
So long as it's okay with @resolvicomai I'd like to incorporate their work particularly around tests included in https://github.com/What-We-Will/community-platform/pull/142. Merging the two PRs together in this way made the most sense to me!

### DIFF
--- a/components/members/MemberFilters.tsx
+++ b/components/members/MemberFilters.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useEffect, useCallback } from "react";
+import { useEffect, useRef, useCallback } from "react";
 import { useRouter, useSearchParams } from "next/navigation";
 import { Input } from "@/components/ui/input";
 import { Label } from "@/components/ui/label";
@@ -22,31 +22,21 @@ interface MemberFiltersProps {
 export default function MemberFilters({ allSkills }: MemberFiltersProps) {
   const router = useRouter();
   const searchParams = useSearchParams();
-  const [searchInput, setSearchInput] = useState(
-    () => searchParams.get("q") ?? ""
-  );
-  const [debouncedSearch, setDebouncedSearch] = useState(
-    () => searchParams.get("q") ?? ""
-  );
 
-  // Sync local state when URL changes (e.g. browser back/forward)
+  const inputRef = useRef<HTMLInputElement>(null);
+  const lastPushedQ = useRef(searchParams.get("q") ?? "");
+  const debounceTimer = useRef<ReturnType<typeof setTimeout> | undefined>(undefined);
+
   useEffect(() => {
     const q = searchParams.get("q") ?? "";
-    // eslint-disable-next-line react-hooks/set-state-in-effect
-    setSearchInput(q);
-    setDebouncedSearch(q);
+    if (q !== lastPushedQ.current) {
+      if (inputRef.current) inputRef.current.value = q;
+      lastPushedQ.current = q;
+    }
   }, [searchParams]);
 
   const skill = searchParams.get("skill") ?? "";
   const referrals = searchParams.get("referrals") === "true";
-
-  // Debounce search input
-  useEffect(() => {
-    const timer = setTimeout(() => {
-      setDebouncedSearch(searchInput);
-    }, SEARCH_DEBOUNCE_MS);
-    return () => clearTimeout(timer);
-  }, [searchInput]);
 
   const updateParams = useCallback(
     (updates: { q?: string; skill?: string; referrals?: string }) => {
@@ -68,24 +58,29 @@ export default function MemberFilters({ allSkills }: MemberFiltersProps) {
     [router, searchParams]
   );
 
-  // Sync URL when debounced search changes
-  useEffect(() => {
-    const currentQ = searchParams.get("q") ?? "";
-    if (debouncedSearch !== currentQ) {
-      updateParams({ q: debouncedSearch });
-    }
-  }, [debouncedSearch, updateParams, searchParams]);
+  const handleSearchChange = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      const value = e.target.value;
+      clearTimeout(debounceTimer.current);
+      debounceTimer.current = setTimeout(() => {
+        lastPushedQ.current = value;
+        updateParams({ q: value });
+      }, SEARCH_DEBOUNCE_MS);
+    },
+    [updateParams]
+  );
 
   return (
     <div className="flex flex-col gap-4 sm:flex-row sm:items-end">
       <div className="flex-1 space-y-2">
         <Label htmlFor="search">Search</Label>
         <Input
+          ref={inputRef}
           id="search"
           type="search"
           placeholder="Search by name, role, or location..."
-          value={searchInput}
-          onChange={(e) => setSearchInput(e.target.value)}
+          defaultValue={searchParams.get("q") ?? ""}
+          onChange={handleSearchChange}
         />
       </div>
       <div className="space-y-2">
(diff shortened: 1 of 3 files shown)


tonyrosario:
Thanks @GeorgeDover!

Overall looks good 👍 Thanks for picking this up, and good diagnosis of the feedback loop. The new architecture is an improvement on the previous (#98) attempt, and on the pre-bug version. Some thoughts before we land it.

## What this PR gets right

- **Resolves the actual root cause of #108.** The pre-PR three-way state sync (`searchInput` ↔ `debouncedSearch` ↔ URL) is what was creating the feedback loop. Reducing to "DOM ref + URL" eliminates it cleanly.
- **Keeps the established conventions** I flagged on #98: URL as source of truth, single `updateParams` helper. (The one place #98 broke from the sibling pattern that #141 preserves.)
- **`useCallback` event handler over `useEffect` debounce.** Agreed — this is the React-recommended shape, and avoids the effect-cascade that bit the previous version.

## What we need to discuss: `router.push` → `router.replace`

This is the **third time in ~4 weeks** a contributor has proposed this swap on MemberFilters (#98 had it too, with a different rationale). It's enough that we should draft an ADR to settle the convention so future contributors don't have to re-litigate it each time.  Let's discuss on Friday (5/22) if you're able to make it!

Short version of where I'm landing:

- Both #68 and #108 are **input-stability bugs**, not back-button bugs. Zero user reports of history pollution; #108's own "Fix" section prescribes "make URL the single source of truth" — which this PR does. The `replace` swap addresses a UX concern that, in this codebase, is so far theoretical.
- [Baymard's user research](https://baymard.com/blog/back-button-expectations) is the strongest external signal: users *expect* `Back` to undo a filter selection. `push` aligns with that expectation; `replace` makes Back skip the page entirely.
- The hybrid pattern (replace while typing, push on commit) is reasonable but is *not* a fix for #108 and shouldn't be bundled with it.

Could we **carve `router.replace` out of this PR** and merge the rest? The input-stability fix is the urgent piece (and your real contribution here); the push-vs-replace question lands cleanly on Friday's meeting agenda (5/22) where I can bring the ADR draft.

## Issues to address regardless of the replace question

Caught these reading the new `handleSearchChange`:

1. **Stale closure on `updateParams`.** The pending timer captures `updateParams` at the time `handleSearchChange` was created. If the URL changes mid-debounce (user toggles `skill` or `referrals` while a search is pending), the timer fires with a stale `searchParams` and overwrites the new filter state. Repro: type "ja", toggle a checkbox before 300ms elapses.

2. **Missing unmount cleanup.** `debounceTimer.current` isn't cleared on unmount. If the user navigates away while a debounce is pending, the timer fires and calls `router.replace` on a no-longer-mounted component.

3. **No tests.** #142 adds a regression test for the input-reset case (`MemberFilters.test.tsx`) — would be nice to fold that in here so we don't regress on the original bug.

## Proposed path

1. **Drop the `router.replace` swap** from this PR (one-line revert).
2. **Add the stale-closure + cleanup fixes** (~10 lines).
3. **Borrow the test from #142** with attribution to @resolvicomai.
4. Merge. We pick up the push/replace question Friday with the ADR draft as the framing doc.

If carving out `replace` is a blocker for you, I'd rather put both #141 and #142 on ice until the ADR lands than merge a swap we'd potentially have to revert. Let me know which path works.

GeorgeDover:
Thanks @tonyrosario! I'll address those last few points for sure! And yes we can definitely remove the change from router.push for now :) I only suggested it because that's the most common way filter inputs are handled elsewhere but I'm not married to it - i'll change it back to how it was and we can defer that discussion for another time!

GeorgeDover:
@tonyrosario no rush on this PR, but just letting you know I've made those changes now so it should be ready whenever anyone has the time for a review :)

tonyrosario:
@GeorgeDover Added a commit to put a pin in this after doing some more research.

- fixed a subtle bug where Enter after a debounce-replace silently no-op'd; split `lastPushedQ` into two refs so commit-dedupe and external-nav-sync stop fighting (full rationale in commit message).
- added unit tests to prove that our changes work and if run against `main` they would fail.
- added e2e test as well but they have to be skipped until we upgrade to Next 16.2.x (https://github.com/What-We-Will/community-platform/pull/137) due a bug (see test for details).

If you like the changes, then the only thing we need to decide is if we want to drop the commits 62cee5a + 6ae83a5.  If yes, either you or me could do it.

I really appreciate the conversation about this, I definitely learned about push/replace/hybrid and will share the research\* so it'll be easier for us all going forward.

\*Research Notes TLDR
- The [official Next.js Learn dashboard tutorial](https://nextjs.org/learn/dashboard-app/adding-search-and-pagination) uses router.replace with useDebouncedCallback for exactly the search-input pattern. 
- replace while typing, push on commit

tonyrosario (approved):

GeorgeDover:
> @GeorgeDover Added a commit to put a pin in this after doing some more research.
> 
> 
> 
> - fixed a subtle bug where Enter after a debounce-replace silently no-op'd; split `lastPushedQ` into two refs so commit-dedupe and external-nav-sync stop fighting (full rationale in commit message).
> 
> - added unit tests to prove that our changes work and if run against `main` they would fail.
> 
> - added e2e test as well but they have to be skipped until we upgrade to Next 16.2.x (https://github.com/What-We-Will/community-platform/pull/137) due a bug (see test for details).
> 
> 
> 
> If you like the changes, then the only thing we need to decide is if we want to drop the commits 62cee5a + 6ae83a5.  If yes, either you or me could do it.
> 
> 
> 
> I really appreciate the conversation about this, I definitely learned about push/replace/hybrid and will share the research\* so it'll be easier for us all going forward.
> 
> 
> 
> \*Research Notes TLDR
> 
> - The [official Next.js Learn dashboard tutorial](https://nextjs.org/learn/dashboard-app/adding-search-and-pagination) uses router.replace with useDebouncedCallback for exactly the search-input pattern. 
> 
> - replace while typing, push on commit
> 
> 

@tonyrosario all sounds good to me! I'm fairly busy today and this weekend so if you have time to push I'd really appreciate it! I think those additions you mentioned are great shoutouts and if they need commit history revision then that's fine by me :)

PR #144, opened 2026-05-24
feat: replace router.push with router.replace

GeorgeDover:
## `router.replace` > `router.push` (for search/filter inputs)
This is a proposal to exchange `router.push` for `router.replace` to provide a more typical back/forward navigation experience for users. Relevant Next.js docs: https://nextjs.org/docs/app/api-reference/functions/use-router

`router.push` records every change to the query params as a separate point in browser history and so you'd have to hit back through all the searches you performed to get to the previous major page. 

Whereas `router.replace` preserves the last input value if you navigated away from the page, but none of the intermediary search values.

---

Note: I do not mind which direction we go - just wanted to document the proposal and get input for future reference!

### DIFF
--- a/components/members/MemberFilters.tsx
+++ b/components/members/MemberFilters.tsx
@@ -63,7 +63,7 @@ export default function MemberFilters({ allSkills }: MemberFiltersProps) {
         if (updates.referrals === "true") params.set("referrals", "true");
         else params.delete("referrals");
       }
-      router.push(`/members${params.toString() ? `?${params.toString()}` : ""}`);
+      router.replace(`/members${params.toString() ? `?${params.toString()}` : ""}`);
     },
     [router, searchParams]
   );


xdhmoore (approved):
Sounds good to me.

timchaffee (approved):
LGTM. I don't have a strong opinion either way, I think this will make it much easier. If folks want a search history we can look into that.

tonyrosario:
After tinkering and doing more research,I think https://github.com/What-We-Will/community-platform/pull/141 is closer to what we want.

kaizengrowth:
Looking more closely at @tonyrosario's comment:

Rather than `router.replace()`,  we mentioned drafting an ADR to standardize the use of `router.push()` in order to keep to UX expectations for users when clicking the back button:
https://github.com/What-We-Will/community-platform/pull/141

Cited in #141 
https://baymard.com/blog/back-button-expectations

Keeping that in mind, I'm going to close this ticket for now. Please let me know if we should raise it again, or discuss this further next Friday for language in an ADR.

PR #148, opened 2026-06-04
docs(adr): add ADR-0004 router.push for filter URLs and add index row

tonyrosario:
Record the filter and search URL convention as it actually shipped: router.push as the default, with a router.replace carve-out during the debounced typing window. Retroactive ADR — the decision shipped in PR #141 (merged 2026-05-31), so status is Accepted 2026-05-31. Implementation credited to @GeorgeDover.


tonyrosario:
will revisit if doc changes can skip the vercel preview

PR #150, opened 2026-06-08
Feature/jest to vitest

tonyrosario:


kaizengrowth (dismissed):

TESTING_STANDARDS.implementation.md
```diff
@@ -152,7 +152,7 @@ All factories live in `lib/__tests__/factories.ts`. Use overrides to vary only w
 
 ## Assertion rules
 
-- Use Jest matchers: `toBe()`, `toEqual()`, `toMatchObject()`
+- Use Vitest matchers (Jest-compatible API): `toBe()`, `toEqual()`, `toMatchObject()`
```
timchaffee: Would we ever want to migrate away from the jest compatible API. Should the end goal be to be 100% vitest native at some point?

timchaffee (dismissed):
Once the security check passes or it's determined it's a non-issue I think this is good to go.

TESTING_STANDARDS.implementation.md
tonyrosario: Fair point and good catch!  Thankfully these are Vitest's native matchers, Jest-shaped names by design so nothing to migrate away from.

I reworded the misleading comment in 50e29aaa21b9f754b05b60d7ebbd714a7336f7e4.  I also scanned for any leftover jest comments and found one here as well 813109f09f960547cab11d1d6fd462d654f67f51, which was also misleading... Vitest actually needs the `@/*` alias and not because 'Jest did it'.

timchaffee (approved):
Lgtm

PR #156, opened 2026-06-22
Fix profile navigation from user avatar

OmegaDeanGrey:
## What changed
- Wrapped the sidebar user avatar/details card in a Link component
- Clicking the user card now navigates to /profile

## Bug
Users could click the personal avatar area but nothing happened.

### DIFF
--- a/app/(app)/app-shell.tsx
+++ b/app/(app)/app-shell.tsx
@@ -253,7 +253,7 @@ export default function AppShell({ children, user }: AppShellProps) {
           </nav>
 
           <div className="sticky bottom-0 z-30 border-t bg-background p-4">
-            <div className="flex items-center gap-3 rounded-lg px-3 py-2">
+            <Link href="/profile" className="flex items-center gap-3 rounded-lg px-3 py-2">
               <UserAvatar
                 avatarUrl={user.avatarUrl}
                 displayName={user.displayName}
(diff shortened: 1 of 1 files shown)


timchaffee (approved):
LGTM

app/(app)/app-shell.tsx
```diff
@@ -253,7 +253,7 @@ export default function AppShell({ children, user }: AppShellProps) {
           </nav>
 
           <div className="sticky bottom-0 z-30 border-t bg-background p-4">
-            <div className="flex items-center gap-3 rounded-lg px-3 py-2">
+            <Link href="/profile" className="flex items-center gap-3 rounded-lg px-3 py-2">
```
tonyrosario: we should an onClick event here:
`onClick={() => setSidebarOpen(false)}`

1. for consistency
2. so that mobile drawer closes when the user taps the card (UX).

Entire line should be:
`<Link href="/profile" onClick={() => setSidebarOpen(false)} className="flex items-center gap-3 rounded-lg px-3 py-2">`

tonyrosario (changes requested):

tonyrosario:
The build failure is due to the workflow orchestration: 
> GitHub deliberately withholds repository secrets from fork-triggered pull_request runs, otherwise any stranger's PR could exfiltrate our secrets. So every ${{ secrets.* }} in preview.yml resolves to an empty string.

@OmegaDeanGrey fix the mobile UX issue from my review and let the job fail for now..  If I can't resolve it quickly, we have other ways to pull in your work as a workaround.

PR #158, opened 2026-06-25
PR 118: Click user profiles in group messages — wrap in `<Link>`

kaizengrowth:
## TLDR

In group conversations, sender avatars and names in message bubbles were not clickable. This wraps them in `<Link>` components so users can navigate to `/members/[userId]`, matching the behavior already used in the 1:1 chat header and group member list.

## Changes

* Added a `SenderProfileLink` helper in `MessageBubble.tsx `that links to the sender's profile when `isGroup` is true
* Wrapped sender avatar and name in both text and file message bubbles
* Passed `isGroup` from `ConversationView` into `MessageBubble`
* 1:1 conversations and own messages are unchanged

## Testing
Open a group conversation with messages from other members:

* When you click on a sender's avatar or name, check that it navigates to their profile page.
* Confirm that file attachment messages also link correctly.

### DIFF
--- a/components/messages/MessageBubble.tsx
+++ b/components/messages/MessageBubble.tsx
@@ -20,11 +20,11 @@ function SenderProfileLink({
   children: ReactNode;
 }) {
   if (!isGroup || !senderId) {
-    return <>{children}</>;
+    return <span className={className}>{children}</span>;
   }
 
   return (
-    <Link href={`/members/${senderId}`} className={className}>
+    <Link href={`/members/${senderId}`} className={cn(className, "hover:underline")}>
       {children}
     </Link>
   );
@@ -117,7 +117,7 @@ function FileMessageBubble({
           <SenderProfileLink
             senderId={senderId}
             isGroup={isGroup}
-            className="text-xs text-muted-foreground mb-1 ml-1 hover:underline"
+            className="text-xs text-muted-foreground mb-1 ml-1"
           >
             {senderName}
           </SenderProfileLink>
@@ -322,7 +322,7 @@ export function MessageBubble({
           <SenderProfileLink
             senderId={message.sender_id}
             isGroup={isGroup}
-            className="text-xs text-muted-foreground mb-1 ml-1 hover:underline"
+            className="text-xs text-muted-foreground mb-1 ml-1"
           >
             {senderName}
           </SenderProfileLink>
--- a/components/messages/MessageBubble.tsx
+++ b/components/messages/MessageBubble.tsx
@@ -1,4 +1,5 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, type ReactNode } from "react";
+import Link from "next/link";
 import { Video, FileText } from "lucide-react";
 import { cn } from "@/lib/utils";
 import { formatRelativeTime } from "@/lib/utils/time";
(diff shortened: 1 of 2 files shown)


timchaffee (dismissed):
LGTM on the surface. Not being super familiar with the code in this area I defer to @tonyrosario. My only thought on the review is there any places for code reuse/deduplication with the 1-1 you mention in the PR "behavior already used in the 1:1 chat header and group member list."

components/messages/MessageBubble.tsx
```diff
@@ -7,6 +8,28 @@ import { getSignedUrl } from "@/lib/storage";
...
+    return <>{children}</>;
+  }
+
+  return (
+    <Link href={`/members/${senderId}`} className={className}>
+      {children}
+    </Link>
+  );
+}
+
```
tonyrosario, line 32: `SenderProfileLink` returns a bare fragment `<>{children}</>` when `!isGroup || !senderId`, silently discarding the `className` prop. Both call sites for the sender name pass `className="text-xs text-muted-foreground mb-1 ml-1 hover:underline"`. 

In every 1:1 conversation where `showSenderInfo` is true, the other person's name renders completely unstyled — losing font size, color, and spacing. The avatar call-site also loses `block rounded-full hover:opacity-80` in the fallback. · 

1. Replace the bare fragment branch (line 23) with `<span className={className}>{children}</span>` so structural Tailwind classes are preserved.
2. Strip `hover:underline` from the call-site className string if interactivity semantics on a span are a concern. (lines 120 and 325)

---
*Generated by `/review-pr-fast` on 2026-06-26. 1 findings validated, 0 false positives filtered. AI_POLICY.md: this automated review assists but does not substitute human review.*

tonyrosario (dismissed):
LGTM!

I think the 1 warning should be fixed to keep UX consistent.

Re: the duplication issue mentioned by @timchaffee and in the PR description, I generally follow the rule of it needs to be duplicated at least 3 times before we consider de-duplicating. The rule that was engrained in me was that it's easier to deal with duplication than the wrong abstraction.

components/messages/MessageBubble.tsx
```diff
@@ -7,6 +8,28 @@ import { getSignedUrl } from "@/lib/storage";
...
+  className,
+  children,
+}: {
+  senderId: string | null | undefined;
+  isGroup: boolean;
+  className?: string;
+  children: ReactNode;
+}) {
+  if (!isGroup || !senderId) {
+    return <span className={className}>{children}</span>;
```
kaizengrowth, line 23: Wrapped content here in `span` to keep styling.

@tonyrosario

components/messages/MessageBubble.tsx
```diff
@@ -81,9 +114,13 @@ function FileMessageBubble({
         )}
       >
         {!isOwn && showSenderInfo && (
-          <span className="text-xs text-muted-foreground mb-1 ml-1">
+          <SenderProfileLink
+            senderId={senderId}
+            isGroup={isGroup}
+            className="text-xs text-muted-foreground mb-1 ml-1"
```
kaizengrowth, line 120: Removed `hover: underline` here.

@tonyrosario

components/messages/MessageBubble.tsx
kaizengrowth: Thanks for your feedback here! 

Now wrapping `{children}` of `SenderProfileLink` in a `span` to preserve Tailwind classes in line 23, and removed `hover: underline` in line 120.

GeorgeDover (approved):

PR #160, opened 2026-06-25
PR 114 - Increase Avatar upload limit from 2MB to 10 MB

kaizengrowth:
## TLDR
Pretty self-explanatory here.

Profile avatar uploads were capped too low and likely rejected many phone photos.

## Question
Is 10 MB enough? Or should we go for a higher cap?

### DIFF
--- a/components/profile/AvatarUpload.tsx
+++ b/components/profile/AvatarUpload.tsx
@@ -6,7 +6,7 @@ import { cn } from "@/lib/utils";
 import { getAvatarColor, getInitials } from "@/lib/utils/avatar";
 import { uploadPublicFile, validateFile } from "@/lib/storage";
 
-const AVATAR_LIMIT_MB = 2;
+const AVATAR_LIMIT_MB = 10;
 const AVATAR_TYPES = ["image/png", "image/jpeg", "image/webp", "image/gif"];
 
 interface AvatarUploadProps {


timchaffee (approved):
LGTM, Supabase is set to 50MB so this is still in those limits. Should we set Supabase to match at 10 for the avatars folder? 

Image
and
Image

tonyrosario:
@kaizengrowth I was NOT expecting this much review depth on this change so I'm going to leave this here so that we can generate a separate issue to deal with this!


## Automated review — pr_116 @ eff7fb5

### ⚠️ Warnings

- **[SECURITY]** components/profile/AvatarUpload.tsx:9 — AVATAR_LIMIT_MB raised from 2 to 10 but validateFile runs entirely client-side; any authenticated user can POST directly to the Supabase Storage REST endpoint, skipping the component and uploading without size ceiling — no file_size_limit is set on the avatars bucket in any migration. · Set file_size_limit on the avatars bucket via a migration (UPDATE storage.buckets SET file_size_limit = 10485760 WHERE id = 'avatars').
  - Related: code-1
  <details>
  <summary>Proposed fix — sensitive — 3 options: see fixes/finding-security-1.md</summary>

  - Option A — Migration + friendly error normalization (recommended):
  - Option B — Migration only:
  - Option C — Migration + upload proxy Route Handler (defer):
  </details>
- **[CODE]** components/profile/AvatarUpload.tsx:9 — The avatars bucket file_size_limit in the Supabase Dashboard may still be 2 MB (matching the old constant); uploads between 2–10 MB now pass client validation but are rejected by Supabase with a raw storage error since handleFile surfaces result.error as-is. · Update the bucket file_size_limit in the Dashboard (or via migration) and document the bucket's configured limit so future constant changes are not split across two places.
  - Related: security-1
  <details>
  <summary>Proposed fix — non-sensitive: see fixes/finding-code-1.md</summary>

  Updating storage.buckets.file_size_limit to 10485760 bytes closes the gap where files between 2-10 MB pass client-side validateFile check but are rejected by Supabase storage with a raw error. The inline comment co-locates the change so future edits to AVATAR_LIMIT_MB prompt updating the migration too.
  </details>
- **[PERF]** components/profile/AvatarUpload.tsx:9 — image/gif is accepted and Supabase image transforms are not applied to lg/xl UserAvatar sizes (UserAvatar.tsx lines 18-22 skip transform params for those sizes), meaning a 10 MB animated GIF is served at full source size in profile pages and group detail views. · Remove image/gif from AVATAR_TYPES (cleanest), or extend UserAvatar.tsx to apply Supabase transform params for lg/xl sizes as well. · [validator-downgraded from CRITICAL]
  <details>
  <summary>Proposed fix — non-sensitive: see fixes/finding-perf-1.md</summary>

  The root fix is removing image/gif from AVATAR_TYPES (see security-2 fix). Additionally, UserAvatar.tsx lg/xl sizes skip Supabase transforms for all types — extend avatarImageUrl guard to cover lg (width=112&height=112) and xl (width=160&height=160) as a separate follow-up.
  </details>

### ❓ Questions

- **[SECURITY]** components/profile/AvatarUpload.tsx:10 — image/gif remains in AVATAR_TYPES and the ceiling just grew to 10 MB; a 10 MB animated GIF is CPU-intensive and could serve as a denial-of-service vector on every page rendering that member's avatar, particularly at lg/xl sizes where Supabase image transforms are not applied.
  - Related: perf-1
## Finding security-2 + perf-1 (consolidated): Remove image/gif from accepted avatar types

  ```diff
  @@ -9,5 +9,5 @@ import { uploadPublicFile, validateFile } from "@/lib/storage";

   const AVATAR_LIMIT_MB = 10;
  -const AVATAR_TYPES = ["image/png", "image/jpeg", "image/webp", "image/gif"];
  +const AVATAR_TYPES = ["image/png", "image/jpeg", "image/webp"];

   interface AvatarUploadProps {
  ```

**Rationale:** Removing image/gif from both the JS validation array and the HTML accept attribute closes the upload gate on animated GIFs, which Supabase storage transforms do not re-encode and which would be served as raw multi-megabyte payloads at lg/xl sizes that bypass the ?width=80&height=80 transform params in UserAvatar.tsx.

---
*Generated by `/review-pr-fast` on 2026-06-26. 4 findings validated, 0 false positives filtered. AI_POLICY.md: this automated review assists but does not substitute human review.*

tonyrosario:
We can merge this PR with one additional change, the review flagged that the file size limit is only client side currently... the system must validate the business logic before storing (and becoming trusted).  In these cases, we want to add business logic and/or a database validation.

So the suggested changes are:
1. Add the file size limit to the database via a migration
2. ~~Add error handling.~~

## 1. Add Supabase Migration

Listing two ways for you

###  Option 1 — Manual (CLI)

1. From the repo root, generate the file (never hand-pick a timestamp) supabase migration new avatars_file_size_limit

2. Open the generated file — it will look like:
  - supabase/migrations/YYYYMMDDHHMMSS_avatars_file_size_limit.sql
  - Add this as the entire body:
```
  -- Enforce the 10 MB avatar upload ceiling server-side.
  -- Matches AVATAR_LIMIT_MB in components/profile/AvatarUpload.tsx.
  UPDATE storage.buckets
  SET file_size_limit = 10485760
  WHERE id = 'avatars';
```

3. Verify locally (optional but recommended)
  supabase db reset

4. Commit
  git add supabase/migrations/
  git commit -m "chore(migrations): set avatars bucket file_size_limit to 10 MB"

  ---
### Option 2 — Ask the agent

  Drop this into Claude Code on the PR branch:

```
Create a Supabase migration that sets file_size_limit = 10485760 on the avatars storage bucket. Use supabase migration new avatars_file_size_limit to generate the file — do not hand-pick a timestamp. The SQL body should be a single UPDATE storage.buckets SET file_size_limit = 10485760 WHERE id = 'avatars' with a comment noting it mirrors AVATAR_LIMIT_MB in components/profile/AvatarUpload.tsx. Follow supabase/migrations/CONVENTIONS.md.
```

The agent will run the supabase migration new command, write the SQL body into the generated file, and can optionally run supabase db reset to verify it applies cleanly before you commit.


## 2. Add Error Handling Logic
My agent actually suggest we handle this in a separate PR/issue.

Current flow in AvatarUpload.tsx:

  1. validateFile runs client-side first (line 40) — if the file is over AVATAR_LIMIT_MB, it returns "File size must be under 10 MB" and stops. That's already a friendly message, already displayed at line 99.
  2. If Supabase rejects the upload anyway, uploadPublicFile in lib/storage.ts:25 passes error.message straight through as a raw string — and that raw string gets rendered at line 99.

  So there are two distinct cases:

  Case 1 — Normal UI user uploads a big file:
  The client-side gate catches it and shows a friendly message. No change needed here — this already works after bumping AVATAR_LIMIT_MB to 10.

  Case 2 — Someone bypasses the UI and POSTs directly to Supabase Storage:
  The bucket limit (once the migration lands) rejects it at the storage layer. But there's no component rendering an error at that point — they're hitting the API directly. The raw error response is fine for that audience.

  The real gap — and this is pre-existing debt, not introduced by this PR — is Case 2b: any situation where the client-side check passes but Supabase rejects for a different reason (transient error, storage quota, network failure). In that case error.message from Supabase gets shown verbatim. That's lib/storage.ts:25 and it affects all buckets, not just avatars.

  Verdict for this PR: No new error handling needed. The client-side gate covers the UX path. The bucket limit is a server-side enforcement backstop, not a UX surface.

  Add to the followup PR:
```
  // lib/storage.ts — normalize raw Supabase storage errors
  if (error) {
    const message = error.message?.toLowerCase() ?? "";
    const friendly = message.includes("exceeded") || message.includes("too large")
      ? `File must be under ${bucket === "avatars" ? "10" : "?"} MB`
      : "Upload failed. Please try again.";
    return { path: "", publicUrl: null, signedUrl: null, error: friendly };
  }
```

  That belongs alongside the gif removal and UserAvatar transform work — it's the same class of debt (pre-existing, not regressions from this PR).

tonyrosario (changes requested):
see comment about adding a DB migration.

kaizengrowth:
Oh wow! Thank you for this review, and notice about client-side / server-side limits. I didn't expect this small change to create such a large issue.

Ok, I'll create the appropriate DB migration.

PR #164, opened 2026-06-26
WARN tracker link added

OmegaDeanGrey:
Added WARN Tracker link to the Resources section of the sidebar.

Added external link to the WARN Tracker Streamlit app
Added Globe icon for visibility
Updated sidebar rendering to support both internal Next.js routes and external links


kaizengrowth (approved):
This is a great fix. I think we were wanting to add this to the front end rather than just in the member platform. I'll make another ticket for that, and add it separately. 

Thanks for adding this!

Issue #165, opened 2026-06-26
Out-of-band admin — Actions fork gate + ruleset required-check + code-owner review


tonyrosario:
## Parent

#157

## What to build

The GitHub admin changes that can't be expressed as code — completed in the UI by a maintainer with admin rights. This is a checklist, not a PR.

> **⚠️ Highest operational risk — do not skip.** GitHub does **not** count a *skipped* required check as passing. After the preview workflow is split, the `preview-deploy` job skips for forks. If that job (or its check) is a required check, fork PRs can **never** merge even after approval. The required check must be **`verify`**, not the deploy job. This bites on the very next fork PR if missed.

## Acceptance criteria

- [ ] Settings → Actions → Fork pull request workflows → **"Require approval for all outside collaborators"** enabled
- [ ] In the `Protect main` ruleset, the **required status check is `verify`** — not the `preview-deploy` job
- [ ] `require_code_owner_review` enabled in the ruleset
- [ ] Branch protection requires the CODEOWNERS review to come from **someone other than the PR author**
- [ ] Verified end-to-end: a fork PR reaches a mergeable state after approval (no perpetually-pending required check)

## Blocked by

- The preview-split issue (the `verify` job must exist before it can be made the required check)
- The CODEOWNERS issue (the file must exist before `require_code_owner_review` has anything to enforce)

Issue #157, opened 2026-06-24
Establish secure fork-contribution intake for CI/CD (no fork previews, execution gate, CODEOWNERS)


tonyrosario:
@timchaffee  This will likely need an ADR but posting this for now for visibility 👀.

# **Decision needed:** how do we accept fork / external-contributor PRs?

**Design rule the solution must hold:** an outside contributor's code must never run in the same workflow step as our secrets, including repo scripts like scripts/ci/*.sh, since a fork can edit those in its PR.   [Example fork PR](https://github.com/What-We-Will/community-platform/pull/156)

I reviewed the situation multiple times and had an independent security reviewer agent come up with this:

### Decision 1 — Do fork PRs get an automatic live preview deploy?
  → No (1a). Forks run secret-free CI only (lint → typecheck → tests → security scan). Deploy/migration steps are skipped for forks; the maintainer pulls the branch locally to verify.
  - Rejected 1b (two-stage workflow_run artifact pipeline): architecturally correct and the documented future path, but easy to implement 90%-correctly with the remaining 10% being a critical exfiltration hole (a single run: bash ./scripts/*.sh in the trusted stage breaks isolation). Not worth the surface area at single-digit-PRs/month.
  - Rejected 1c (Vercel native Git integration): the "authorize deployment" button is approval UX, not a code-isolation boundary — next build still executes fork code (next.config.js, build plugins, postinstall). Creates a false sense of safety, and leaves the Supabase preview-DB step homeless.

### Decision 2 — How aggressively to gate workflow execution on fork PRs?
  → Require approval for all outside collaborators (2a). No CI runs until a maintainer clicks "Approve and run."
  - Rejected 2b (first-time contributors only): once a contributor has one merged PR, GitHub treats them as "known" and all later PRs bypass the gate — even though provenance can shift PR-to-PR (especially with
  AI-assisted contributions). At our volume the friction of gating everyone is negligible.

### Decision 3 — CODEOWNERS scope?
  → Broad (3a), expanded. Security-critical paths: supabase/migrations/, .github/, scripts/ci/, lib/supabase/, middleware.ts, docs/adr/, plus package.json and the lockfile (see solution step 3). Also gates
  maintainers' own PRs on these paths — accepted tradeoff at this volume.
  - Rejected 3b (docs/adr only): that's documentation governance, not a security control.
  - Rejected 3c (general "require 1 review"): doesn't distinguish a copy change from a migration that drops a column.

[...]

PR #167, opened 2026-06-27
ci(preview): split into secret-free verify + fork-gated preview-deploy

### LINKED ISSUE #161: ADR — fork-contribution intake posture
## Parent

#157

## What to build

An Architecture Decision Record capturing the fork-contribution intake decision so it's discoverable and the rationale survives. It records:

- The **three-gate model**: what's allowed to *run* (fork CI execution), what's allowed to *merge* (path ownership), and the hard rule that **no secret ever shares a workflow step with fork-controlled code** (including `scripts/ci/*`).
- **Decisions 1–4** with the chosen option, the rejected alternatives, and why (no fork previews; require approval for all outside collaborators; broad CODEOWNERS incl. `package.json` + lockfile; drop the preview-DB migration for forks).
- The **deferral of the secure `workflow_run` preview pipeline (option 1b)** with a concrete revisit trigger — without a recorded threshold this silently never gets reconsidered.
- The baseline security audit result.

## Acceptance criteria

- [ ] ADR added under `docs/adr/` following the project template and README status rules
- [ ] Three-gate model documented (run / merge / secret isolation) with the hard invariant stated explicitly
- [ ] Decisions 1–4 recorded with chosen option, rejected alternatives, and rationale
- [ ] Deferral of option 1b recorded with an explicit revisit trigger: external PR volume exceeds ~10/month, **or** an external contributor needs a shareable preview URL
- [ ] `pull_request_target` audit result noted as baseline: clean, no exploitable trigger present (2026-06-26)
- [ ] Status set per ADR conventions; index row added

## Blocked by

None — can start immediately. Soft predecessor to the contributor-onboarding issue (it documents the dispositions); the code issues (preview split, CODEOWNERS) may proceed in parallel since the decision is already made.

tonyrosario:
# Layer 1 — split preview workflow into secret-free `verify` + fork-gated deploy

Closes #161 (parent #157). Full rationale and acceptance criteria are in the issue.

Splits `preview.yml` so untrusted fork-PR code can never share a job with a secret:

- **`verify`** — secret-free (lint, typecheck, tests, migration-collision gate, paths-filter). Runs for every PR including forks.
- **`security-scan`** — secret-free. Unchanged; still its own always-runs job.
- **`preview-deploy`** — every secret-bearing step (Supabase + Vercel + Slack). Gated at the job level to same-repo PRs via one `if:`; skipped entirely for forks.

## What to review
- The job-level fork gate on `preview-deploy` (`head.repo.full_name == github.repository`) — one condition, no per-step secret gating.
- That no `secrets.*` reference leaked into `verify` or `security-scan`.
- `needs: verify` + `needs.verify.outputs.migrations` — keeps the paths-filter secret-free while still driving the Supabase push.

An independent GitHub Actions security review confirmed the isolation holds (gate non-spoofable, no `${{ github.event.* }}` injection sinks, actions SHA-pinned).

## Not obvious from the diff
- **Required check must become `verify` after this merges** — admin step, tracked in #165. `preview-deploy` must never be required (it's skipped for forks → would block every fork PR). Sequenced after merge so the ruleset doesn't require a check that doesn't exist yet.
- **Minor behavior change:** the Slack *failure* notice now fires only for deploy-stage failures, not lint/test failures — that step carries a secret so it can't live in a fork-reachable job. Verify failures surface as the failing `verify` check instead.

## Follow-ups (non-blocking, pre-existing)
- `npm install -g vercel@…` runs without `--ignore-scripts` in the deploy job (also in `production.yml`) — worth pinning vercel as a committed dev-dependency in a separate PR.
- The live fork-PR test can't run pre-merge; recorded as a post-merge check.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,5 +1,13 @@
 name: PR Preview Vercel Deploy with Supabase preview DB
 
+# PR preview deploy. Split into three jobs so untrusted fork code can never share
+# a job with a secret.
+#
+# INVARIANT: every `secrets.*` reference lives in preview-deploy, which is gated off
+# for forks by a single job-level `if:`. Keep verify and security-scan secret-free
+# (they run fork code), and don't move a secret-bearing step — or a `run:` of a repo
+# script — out of preview-deploy.
+
 on:
   pull_request:
     types: [opened, synchronize, reopened]
@@ -11,14 +19,20 @@ permissions:
   contents: read
 
 jobs:
-  pr-preview:
+  # Secret-free checks; runs for every PR, forks included, so fork code is safe here.
+  # The migrations output is consumed by preview-deploy.
+  # Required check: the branch-protection ruleset must point at this job. If you
+  # rename it, update the ruleset — preview-deploy never runs for forks, so the
+  # required check must be one that always runs.
+  verify:
     if: github.event.action != 'closed'
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     permissions:
       contents: read
-      pull-requests: write
+    outputs:
+      migrations: ${{ steps.changes.outputs.migrations }}
 
-    timeout-minutes: 10
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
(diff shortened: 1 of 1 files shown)


GeorgeDover (approved):
Looks great!

xdhmoore:
I had a little trouble understanding this one, but it looks good to me, best I can tell. A couple questions:

1. Will this allow preview env's to be spun up after a 3rd party's PR is merged?
2. Are there any integration tests that use secrets that will be affected by this?
3. Will someone be able to spin up their own 3rd party vercel instances with this? For example if they issue a PR to themselves?

tonyrosario:
Sorry @xdhmoore - I missed this comment... Hopefully I tuned my notifications properly to see this in a more timely manner.

> 1. Will this allow preview env's to be spun up after a 3rd party's PR is merged?

No, merging doesn't grant anything automatically. Once a PR merges, that code has been reviewed and becomes part of main. The next PR from a collaborator will generate a preview normally.

> 2. Are there any integration tests that use secrets that will be affected by this?

No.  We're just splitting/renaming the job here (line 27, renamed 'pr-preview' to 'verify')

After doing a quick search, it seems that only `lib/utils/get-site-url.test.ts`  uses VERCEL_URL/NEXT_PUBLIC_SITE_URL and even then, it's just fake strings.

`npm test` runs in **verify** and has 0 secret references.


> 3. Will someone be able to spin up their own 3rd party vercel instances with this? For example if they issue a PR to themselves?

Had to lean on AI for some additional details but the TLDR is:
_No. Repo secrets are never handed to fork-triggered runs in the first place, and this PR adds an explicit job-level gate that skips the deploy job entirely for forks, so there's no path to a Vercel deploy with our credentials._

### Details for Q3
There are actually two independent layers doing that work:
#### Layer 1 - GitHub platform-level, not from this diff
Repo secrets are scoped to the repository — they only get injected into workflow runs that GitHub Actions considers to belong to this repo's trust boundary. For a pull_request event (which is what this workflow uses, not pull_request_target), GitHub explicitly withholds all secrets.* from fork-originated runs regardless of what the workflow file says.

So even before this PR's fork gate existed, a fork PR's steps referencing secrets.VERCEL_TOKEN would have resolved to an empty string — they never had the actual token value to begin with.

#### Layer 2 - this diff, explicit and job-scoped
The if: github.event.pull_request.head.repo.full_name == github.repository gate on preview-deploy means the job doesn't even attempt to run for fork PRs — it's skipped outright, before checkout, before any step executes.

PR #183, opened 2026-07-01
On local, allow engineers to create new users as Admin role without need for approval to platform

### LINKED ISSUE #182: Remove deprecated Supabase variables from `env.example` template
## TLDR
Supabase has deprecated ANON_KEY and SERVICE_ROLE_KEY, so these no longer need to be in our `env.example` file, and may confuse folks onboarding.

kaizengrowth:
## Summary

- Adds an opt-in `LOCAL_DEV_AUTO_ADMIN` env flag so new volunteer engineers can finish onboarding locally without manual Supabase approval.
- When enabled, onboarding sets `approval_status: approved` and `role: admin`, and skips the admin notification email.
- Documents the flag in `.env.example` and the README getting-started section.

Production behavior is unchanged unless the flag is explicitly set.

## Test plan

- [ ] With `LOCAL_DEV_AUTO_ADMIN=true` in `.env.local`, sign up and complete onboarding → lands on `/dashboard` (not `/pending-approval`) with admin access (e.g. `/admin/approvals` loads).
- [ ] Without the flag (or set to anything other than `"true"`), complete onboarding → `approval_status` is `pending` and user is held at `/pending-approval`.
- [ ] With the flag enabled, confirm no admin notification email is sent on onboarding.
- [ ] Confirm preview/production deploys do **not** set this env var.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,8 @@ Open [http://localhost:3000](http://localhost:3000). Sign up via the auth flow;
 
 To manually approve yourself, open Supabase > Table Editor > "profiles". You should see a row for your user. Change the "approval_status" column from "pending" to "approved".
 
+For local volunteer setup, add `LOCAL_DEV_AUTO_ADMIN=true` to `.env.local` to skip manual approval and grant admin on first onboarding.
+
 After onboarding you’ll see the main app (dashboard, events, groups, messages, members, profile).
 
 
--- a/.env.example
+++ b/.env.example
@@ -9,6 +9,8 @@ SUPABASE_SERVICE_ROLE_KEY=
 
 # ── App ───────────────────────────────────────────────────────────────────────
 NEXT_PUBLIC_SITE_URL=http://localhost:3000
+# Local dev only — auto-approve and grant admin on first onboarding
+LOCAL_DEV_AUTO_ADMIN=true
 
 # ── Email (Nodemailer / Gmail) ────────────────────────────────────────────────
 GMAIL_USER=
--- a/app/onboarding/actions.ts
+++ b/app/onboarding/actions.ts
@@ -34,6 +34,8 @@ export async function completeOnboarding(
     return { error: "LinkedIn URL is required to verify your background." };
   }
 
+  const localDevAutoAdmin = process.env.LOCAL_DEV_AUTO_ADMIN === "true";
+
   const { error } = await supabase.from("profiles").upsert(
     {
       id: user.id,
@@ -47,7 +49,9 @@ export async function completeOnboarding(
       linkedin_url: data.linkedin_url || null,
       timezone: safeTimezone(data.timezone),
       is_onboarded: true,
-      approval_status: "pending",
+      ...(localDevAutoAdmin
+        ? { approval_status: "approved" as const, role: "admin" as const }
+        : { approval_status: "pending" as const }),
     },
     { onConflict: "id" }
   );
@@ -56,12 +60,14 @@ export async function completeOnboarding(
     return { error: error.message };
   }
 
-  // Notify admin that a new account needs review
-  await notifyAdminOfNewApplication({
-    displayName: data.display_name,
-    linkedinUrl: data.linkedin_url,
-    userEmail: user.email ?? "unknown",
-  });
+  if (!localDevAutoAdmin) {
+    // Notify admin that a new account needs review
+    await notifyAdminOfNewApplication({
+      displayName: data.display_name,
+      linkedinUrl: data.linkedin_url,
+      userEmail: user.email ?? "unknown",
+    });
+  }
 
   revalidatePath("/", "layout");
   return {};


GeorgeDover (approved):

tonyrosario (changes requested):
🔴 HOLD please.

There does seem to be a security issue that we have to address but before making any changes, I want to take a closer look at the requirements to double-check that there isn't another way to accomplish the goal.

👀 What I can say that from my experience, using an **environment conditional** to change security/authorization behavior like grant privileges, bypass an approval gate, weaken a check, inside a shared production code path is the wrong tool for the job.

- Environment conditionals are fine for configuration differences. They are the wrong tool for provisioning or privilege
- A dev-only privileged shortcut should live somewhere that cannot ship to production at all, not as a branch in the request handler guarded by a runtime string.
  - Example options: seed data, fixtures, or a local-only script/task.

## Automated review — supabase-local-fix
Reviewed at branch tip `4fb26a4` · full diff `d3ee43c..4fb26a4`

One 🚫 blocker and one ⚠️ warning — both auth/RLS-sensitive, and they **compound**: the blocker is the application-layer hole, the warning is the database-layer hole sitting behind it. Fixing only the blocker still leaves a direct-API self-elevation path.

### 🚫 Blocker (must resolve before merge)

**`LOCAL_DEV_AUTO_ADMIN` has no `NODE_ENV` guard → production privilege escalation** — `app/onboarding/actions.ts:37`

The flag is checked with no `NODE_ENV !== "production"` guard. `next build` sets `NODE_ENV=production` on **every** Vercel deploy (production *and* preview), so if this var is ever set in a hosted environment, every user who completes onboarding is silently granted `role:"admin"` + `approval_status:"approved"`. The anon-key SSR client honors RLS, but the `profiles` UPDATE policy has no column restriction (see the warning below), so the upsert succeeds. `role='admin'` gates privileged RLS across ~10 tables — full escalation with no server-side backstop. Flagged independently by both the security and code reviewers.

<details><summary>Recommended fix — environment guard + strip the admin grant (least-privilege)</summary>

Gate the branch on `NODE_ENV` *and* remove `role:"admin"` from the auto-grant so local convenience is scoped to approval-bypass only (the actual DX need):

```diff
// app/onboarding/actions.ts
-  const localDevAutoAdmin = process.env.LOCAL_DEV_AUTO_ADMIN === "true";
+  // NODE_ENV is "production" in all `next build` outputs (Vercel prod, preview, any hosted deploy).
+  // This guard fails closed on every non-local surface.
+  const localDevAutoAdmin =
+    process.env.NODE_ENV !== "production" &&
+    process.env.LOCAL_DEV_AUTO_ADMIN === "true";

     ...(localDevAutoAdmin
-    ? { approval_status: "approved" as const, role: "admin" as const }
+    ? { approval_status: "approved" as const }
       : { approval_status: "pending" as const }),
```

```diff
// README.md
-For local volunteer setup, add `LOCAL_DEV_AUTO_ADMIN=true` to `.env.local` to skip manual approval and grant admin on first onboarding.
+For local volunteer setup, add `LOCAL_DEV_AUTO_ADMIN=true` to `.env.local` to skip manual approval (sets `approval_status=approved` on first onboarding).
+To also gain admin privileges locally, open Supabase > Table Editor > profiles and set `role=admin` on your row.
```

**Alternatives:** (B) remove `LOCAL_DEV_AUTO_ADMIN` entirely and document the Table-Editor path — strongest, removing code beats guarding it; (C) `!process.env.VERCEL_ENV` guard — rejected: depends on platform-specific env injection and silently re-enables off-Vercel. `NODE_ENV` is a framework-guaranteed signal, so it's the robust backstop.
</details>

### ⚠️ Warning — `profiles` RLS allows self-write of `role`/`approval_status`

`supabase/migrations/001_profiles.sql:37` — the "Users can update own profile" policy is `WITH CHECK (id = auth.uid())` with **no column-level restriction**, so any authenticated user can `update({ role: 'admin' })` on their own row via a direct PostgREST call. Pre-exists the diff, but PR-183's anon-key `role:"admin"` upsert makes it load-bearing — a crafted direct call could self-elevate **even without the env var**. Postgres `WITH CHECK` evaluates the NEW row as a whole and can't compare against OLD values, so this must be enforced at the column or trigger layer.

<details><summary>Recommended fix — BEFORE UPDATE trigger on the protected columns</summary>

New **timestamp-named** migration (per ADR-0002). The codebase already has two `BEFORE UPDATE` triggers on `profiles`/`groups`, so the pattern is established:

```sql
-- supabase/migrations/YYYYMMDDHHMMSS_protect_profiles_role_columns.sql
CREATE OR REPLACE FUNCTION public.guard_profile_protected_columns()
RETURNS TRIGGER AS $$
BEGIN
  -- Service role bypass: the admin approval flow uses createServiceClient().
  IF auth.role() = 'service_role' THEN
    RETURN NEW;
  END IF;
  IF (OLD.role IS DISTINCT FROM NEW.role) AND (OLD.role <> 'admin') THEN
    RAISE EXCEPTION 'permission denied: role is a protected column';
  END IF;
  IF (OLD.approval_status IS DISTINCT FROM NEW.approval_status) AND (OLD.role <> 'admin') THEN
    RAISE EXCEPTION 'permission denied: approval_status is a protected column';
  END IF;
  RETURN NEW;
END;
$$ LANGUAGE plpgsql;

CREATE TRIGGER protect_profile_protected_columns
  BEFORE UPDATE ON public.profiles
  FOR EACH ROW
  EXECUTE FUNCTION public.guard_profile_protected_columns();
```

**Companion app change:** the local-dev `role`/`approval_status` write must move to a `createServiceClient()` call (mirroring `app/admin/approvals/actions.ts:42`), not the anon-key upsert — otherwise the trigger rejects it.

**Alternatives:** (B) `REVOKE UPDATE (role, approval_status) ON public.profiles FROM authenticated` — simplest, but an invisible constraint that a later broad `GRANT` can silently supersede (triggers survive re-grants). (C) defer + ADR first.

**Decision to record:** this is the codebase's first column-level protection on a security-critical table — worth an ADR (`0006-profiles-protected-column-enforcement`) settling trigger-vs-privilege and the canonical privileged-write path. Not a blocker; Option A ships independently.
</details>

---
*Generated by `/review-pr-fast` on 2026-07-02. 2 findings validated, 0 false positives filtered. AI_POLICY.md: this automated review assists but does not substitute human review.*

tonyrosario:
Pasting another review after doing an independent analysis against the codebase (not just the diff) and the PR description.  Posting so that we can start from here later.

I still need to read the entire thing but I wanted to get feedback on other PR's before our next meeting!

## Review — supabase-local-fix (`LOCAL_DEV_AUTO_ADMIN`)
Reviewed at branch tip `4fb26a4` · full diff `d3ee43c..4fb26a4`


## TL;DR

**Recommendation: don't merge this approach — relocate the effort.** The goal ("a local contributor lands as an approved admin without hand-editing Supabase") is a **provisioning/tooling** concern. This PR solves it by adding a privilege-granting branch to the **production onboarding request path** (`app/onboarding/actions.ts`, a `"use server"` action that runs for *every* real signup), gated only by an env var with no `NODE_ENV`/`VERCEL_ENV` backstop. The fix isn't to guard that branch better — it's to move "provision a local admin" somewhere that **cannot ship to production at all**.

## The goal (Y), stripped of the mechanism

From the PR summary: *new volunteer engineers should finish onboarding locally without manual Supabase approval, and have admin so they can exercise admin features.* The real need is **"cheaply provision a known approved-admin user in my own dev database."** Teaching the onboarding flow to grant admin is incidental to that need — and it's the part that creates the risk.

## Why the shape is wrong (the principle)

Environment conditionals are the right tool for **configuration** (log levels, base URLs, mock vs. real mailer). They're the wrong tool for **authorization/provisioning**. The reason:

> If a code path exists in the production build, it can run in production. An `if (LOCAL_DEV_AUTO_ADMIN)` branch that grants admin is **dormant, not absent** — one leaked or mistyped env var and it fires. "Dead in prod by convention" is not a security boundary.

Two things in the PR make this concrete:

- **"Production behavior is unchanged unless the flag is explicitly set"** treats the *absence* of an env var as the security boundary. It isn't one — it's a default that any misconfiguration flips.
- The test plan's *"Confirm preview/production deploys do **not** set this env var"* is a **manual checkbox** standing in for an enforced guard. If the safety of an admin-grant depends on a human remembering not to set a variable, the design has already failed closed-vs-open the wrong way.
- Adding `LOCAL_DEV_AUTO_ADMIN` to **`.env.example`** (outside the reviewed diff — `.env*` is excluded) *advertises* the switch to every environment, nudging it toward exactly the copy-paste leak above.

If the flag is ever set in a hosted env, every onboarding user silently becomes admin across the ~10 tables whose RLS keys on `role='admin'`, and the admin-notification email is suppressed so no one notices. That's a fail-open blast radius for a problem with zero reason to exist in production code.

## What to do instead — provision out of the request path

Leave `app/onboarding/actions.ts` untouched. Move the admin provisioning into tooling that can't run in prod. In rough order of preference for this repo:

**Option A — documented SQL promotion step (recommended, lowest cost).** Extend the workflow the repo already teaches (`e2e/README.md` provisions test state with `UPDATE public.profiles SET …`; `README.md:86` already documents the manual approve step). Add `role='admin'` to it:

```sql
-- Run against your own local/dev Supabase project after signing up:
UPDATE public.profiles
SET approval_status = 'approved', role = 'admin'
WHERE id = (SELECT id FROM auth.users WHERE email = 'you@example.com');
```

One copy-paste per fresh DB (not per signup, if you reuse your dev user), zero production surface, zero new code paths, and it matches an existing convention.

**Option B — a `scripts/dev/make-admin` helper (fallback, if the SQL is too much friction).** A small npm task that promotes by email using the existing sanctioned service-role client (`lib/supabase/service.ts` / `createServiceClient()`, the same primitive `app/admin/approvals/actions.ts` already uses for privileged writes). Still a *deliberate, developer-invoked* action — categorically safer than an automatic request-path branch.

**Not `supabase/seed.sql` (for now).** It looks like the "correct" answer, and `config.toml` even wires the hook (`[db.seed] sql_paths = ["./seed.sql"]`), but there's no seed file and the hook only fires on `supabase db reset` (local Docker). Per ADR-0002, local dev here runs against a **personal cloud project via `db push`**, so `seed.sql` would never execute in the current workflow. It becomes the right answer *if/when* the team adopts `supabase start` — worth revisiting then. (If some contributors already run Docker locally, tell me and I'll add it as a first-class path.)

## Two issues to fix regardless of this PR

These are independent of the DX decision and should be tracked on their own:

1. **🔴 RLS self-elevation (real vulnerability, latent today).** `supabase/migrations/001_profiles.sql` — the `profiles` UPDATE policy is `WITH CHECK (id = auth.uid())` with **no column restriction and no trigger backstop**, so *any* authenticated user can `update({ role: 'admin', approval_status: 'approved' })` on their own row via a direct PostgREST call — env var or not. This PR leans on that hole (the anon-key upsert only works *because* of it). Close it with a `BEFORE UPDATE` trigger (the codebase already has two on `profiles`/`groups`) or a column-level `REVOKE`, and record the decision as an ADR — it's the first column-level protection on a security-critical table. Note this fix would *correctly break* the submitted approach, which is a signal the approach was standing on a bug.

<details><summary>Trigger sketch (timestamp-named migration per ADR-0002)</summary>

```sql
CREATE OR REPLACE FUNCTION public.guard_profile_protected_columns()
RETURNS TRIGGER AS $$
BEGIN
  IF auth.role() = 'service_role' THEN RETURN NEW; END IF;  -- admin approval flow uses createServiceClient()
  IF (OLD.role IS DISTINCT FROM NEW.role) AND (OLD.role <> 'admin') THEN
    RAISE EXCEPTION 'permission denied: role is a protected column';
  END IF;
  IF (OLD.approval_status IS DISTINCT FROM NEW.approval_status) AND (OLD.role <> 'admin') THEN
    RAISE EXCEPTION 'permission denied: approval_status is a protected column';
  END IF;
  RETURN NEW;
END;
$$ LANGUAGE plpgsql;

CREATE TRIGGER protect_profile_protected_columns
  BEFORE UPDATE ON public.profiles
  FOR EACH ROW EXECUTE FUNCTION public.guard_profile_protected_columns();
```
</details>

2. **Schema/app drift.** `001_profiles.sql:17` defaults `approval_status` to `'approved'`, while onboarding writes `'pending'`. The approval gate lives only in app code, not the schema. Worth aligning so the DB default is prod-safe on its own.

---
*Design review + `/review-pr-fast` findings, 2026-07-02. Independent design analysis (opus) + validator-confirmed security findings. AI_POLICY.md: this automated review assists but does not substitute human review.*

kaizengrowth:
@tonyrosario Ah, this is really insightful, and helpful to know! I didn't realize this can also be bypassed in prod. Ok, I will first try to implement Option A, utilizing steps in the `e2e/README.md` file.

kaizengrowth:
@tonyrosario Thanks again for the thorough review. The `env-flag` approach was the wrong shape here.

I went with **Option A**:
- Removed `LOCAL_DEV_AUTO_ADMIN` from onboarding / `.env.example` (no privilege grant in the production request path)
- Documented local approve + optional admin promotion via Supabase SQL / Table Editor in the README, including a short setup checklist and the existing profile-grant troubleshooting

Leaving the RLS protected-column fix and `approval_status` default drift as separate follow-ups, as you suggested, and opened a separate issue here for this:
https://github.com/What-We-Will/community-platform/issues/197#issue-4857036960

Could you take another look when you have a chance?

xdhmoore:
@kaizengrowth the file comparison tab shows a lot of unrelated changes for this branch. Could you do a rebase?

README.md
```diff
@@ -81,12 +83,57 @@ npx supabase link
...
+  SELECT id FROM auth.users WHERE email = 'you@example.com'
+);
+```
 
-To manually approve yourself, open Supabase > Table Editor > "profiles". You should see a row for your user. Change the "approval_status" column from "pending" to "approved".
+- Keep `role = 'admin'` if you need to exercise admin features (e.g. `/admin/approvals`).
+- Omit `role = 'admin'` (or set `role = 'member'`) if you only need membership approval.
 
-After onboarding you’ll see the main app (dashboard, events, groups, messages, members, profile).
+**Option 2 — Table Editor:** Supabase Dashboard → **Table Editor** → `profiles` → your row → set `approval_status` to `approved`, and set `role` to `admin` only if you need admin UI.
```
xdhmoore, line 134: Optional:
Might as well remove option 2 to make things simpler to follow.

README.md
```diff
@@ -81,12 +83,57 @@ npx supabase link
...
+Do **not** put privilege grants in app code or env flags — promote via SQL or the Table Editor only.
+
+#### Troubleshooting: `permission denied for table profiles`
+
+Only needed if step 1 fails. Your Supabase project may be missing grants for the `authenticated` role.
+
+In the Supabase Dashboard → **SQL Editor**, check permissions:
+
+```sql
+select
```
xdhmoore, line 102: Optional:
since GRANT is idempotent, i think technically they don't need to run this first statement. They can just run the GRANT to streamline things.

xdhmoore: Optional:
It would streamline also if you just put the GRANT run above in step 2 instead of only running it when something fails.

README.md
```diff
@@ -81,12 +83,57 @@ npx supabase link
 npm run dev
 ```
 
-Open [http://localhost:3000](http://localhost:3000). Sign up via the auth flow;
+Open [http://localhost:3000](http://localhost:3000), then follow this local setup checklist:
+
+1. Sign up and complete onboarding. New accounts start as `approval_status = pending` (same as production).
+2. If onboarding fails with `permission denied for table profiles`, fix grants (see troubleshooting below), then retry onboarding.
+3. Promote your user to approved (and admin if you need admin UI) in **your own** local/dev Supabase project.
```
xdhmoore, line 90: Optional:
It's not clear to me here that "Promote your user" is actually referring to the SQL below. My vote would be to include the SQL up here instead of below In general, I think in this doc having all the steps in chronological order would be clearer then trying to separate out those that may be skippable. My two cents.

README.md
```diff
@@ -81,12 +83,57 @@ npx supabase link
...
+Then retry onboarding.
+
+#### Promote your local user (approve + optional admin)
+
+Use this after onboarding succeeds (step 3 above).
+
+**Option 1 — SQL Editor (recommended):** Supabase Dashboard → **SQL Editor**. Replace the email with the one you signed up with:
+
+```sql
+UPDATE public.profiles
```
xdhmoore, line 123: Optional:
is there a way we can use the supabase cli tool to run these sql commands either automatically or at user invocation of a script instead of making them find the sql interface in the supabase website and run them manually from there? Then again it's probably a good learning point to know where the sql interface is in the website.

kaizengrowth:
> @kaizengrowth the file comparison tab shows a lot of unrelated changes for this branch. Could you do a rebase?

Ah, sorry. Thanks for spotting this, @xdhmoore! Will do.

tonyrosario:
just to confirm - @kaizengrowth was the rebase completed?

kaizengrowth:
> just to confirm - @kaizengrowth was the rebase completed?

Yes, it is now @tonyrosario

PR #187, opened 2026-07-02
feat(onboarding): allow GitHub or website signup link instead of LinkedIn

### LINKED ISSUE #170: Allow platform signup with GitHub or personal website link as alternative to LinkedIn
### Problem
A member requested the ability to join the platform without providing a LinkedIn profile link. Currently LinkedIn appears to be a required (or only-supported) field for identity/profile verification at signup.

### Request
Allow members to be added to the platform using a GitHub profile URL or other personal website link as an alternative to a LinkedIn link.

### Context
Not everyone maintains a LinkedIn profile — some members (e.g. developers, organizers) may prefer to link a GitHub profile or personal site instead. This is a barrier to entry for an otherwise-eligible member.

### Suggested approach

* Accept one of: LinkedIn URL, GitHub URL, or personal website URL (instead of requiring LinkedIn specifically)
* Update signup/profile form validation accordingly
* Update any onboarding copy that currently references "LinkedIn" specifically

**Reported by:** member request by Brian Kay

GeorgeDover:
## Summary
Addresses https://github.com/What-We-Will/community-platform/issues/170

- Onboarding now accepts LinkedIn, GitHub, or a personal website as the verification link. Previously LinkedIn was the only option and was required. At least one of the three is still required.
- Admin notification emails and the `/admin/approvals` review page show whichever link(s) an applicant actually provided, instead of assuming LinkedIn.
- Updates onboarding, approvals, and pending-approval copy to no longer reference LinkedIn specifically.
- No DB migration needed; `github_url`/`portfolio_url` columns already existed on profiles.

## Test plan

> Testing this is a bit cumbersome due to needing to jump between personas (new user, admin). Someone more familiar with admin view would be preferable for a review as they will know better what has changed visually!

- [ ] `app/onboarding/actions.test.ts`: `completeOnboarding` errors when all three links are empty, succeeds with any single link or combination.
- [ ] `app/onboarding/onboarding-form.test.tsx`: form renders three optional inputs, blocks submit with none filled, submits with only one filled.
- [ ] `npx tsc --noEmit`, `eslint`, and full `npm test` suite pass.
- [ ] Manually verify in browser: submit onboarding with only a GitHub link → lands on `/pending-approval`; `/admin/approvals` shows the GitHub link; admin notification email lists it.


app/onboarding/onboarding-form.tsx
```diff
@@ -62,6 +68,12 @@ export default function OnboardingForm({
     setError(null);
     setLoading(true);
 
+    if (!linkedinUrl.trim() && !githubUrl.trim() && !portfolioUrl.trim()) {
+      setError("Provide at least one link so we can verify your background.");
```
kaizengrowth: nit:  the error on the client side doesn't have the same wording as the one on the server side ("Please provide at least one of LinkedIn, GitHub, or a personal website so we can verify your background.") -- actions.ts, line 42 

This is ok. Just noting.

app/onboarding/onboarding-form.tsx
```diff
@@ -205,21 +219,44 @@ export default function OnboardingForm({
...
-              type="url"
-              placeholder="https://linkedin.com/in/username"
-              value={linkedinUrl}
-              onChange={(e) => setLinkedinUrl(e.target.value)}
-              required
-            />
-            <p className="text-xs text-muted-foreground">
-              Required to verify your background as a tech worker
-            </p>
+          <div className="space-y-3 rounded-md border p-4">
```
kaizengrowth, line 311: Really like how you thoughtfully grouped verification here in one section with a shared explanation.

tonyrosario:
## Automated review — feat/170-github-or-website-signup-link
Reviewed at branch tip `6433923` · full diff `e488fef..6433923` (7 files, 560 lines)

There are **two distinct fixes** to make here (a third finding shares the second fix, and there's one trivial test nit).

### 🚫 Blocker (must resolve before merge)

**HTML injection in the admin notification email** — `app/onboarding/actions.ts`

User-supplied values are interpolated straight into a nodemailer HTML template with no escaping. **This is broader than the URLs** — an independent security pass (verified against the source) found that `displayName` and `userEmail` are *also* raw sinks in both the HTML body and the `subject:` line. Escaping only `link.url` does **not** close the hole: a `display_name` of `` still injects into the admin's mail client.

Unescaped user-controlled sinks (all must be fixed):
- `${displayName}` — HTML body **and** `subject:` header
- `${userEmail}` — HTML body
- `${link.url}` (×2: `href` and link text)

**Verified:** no reusable HTML-escape utility exists in the codebase. The only escaping helper, `lib/seo.ts:27` (`.replace(/</g, "\\u003c")`), is JSON-LD-specific — it escapes only `<` for `<script>` safety and does not handle `"`, `>`, `&`, or `'`. So a new util is the right call.

<details><summary>Recommended fix — add a reusable <code>escapeHtml</code> utility (+ test) and apply it to EVERY sink</summary>

**Placement:** `lib/utils/` uses flat files with a paired `*.test.ts` (see `format.ts`/`format.test.ts`, `time.ts`/`time.test.ts`). Follow that convention — `lib/utils/html.ts` **plus** `lib/utils/html.test.ts`, not a nested `html/` dir.

```diff
// lib/utils/html.ts (new file)
+export function escapeHtml(str: string): string {
+  return str
+    .replace(/&/g, '&amp;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#x27;');
+}

// app/onboarding/actions.ts
+import { escapeHtml } from '@/lib/utils/html';

   // subject line
-      subject: `[New Application] ${displayName} is requesting membership`,
+      subject: `[New Application] ${escapeHtml(displayName.replace(/[\r\n]/g, ''))} is requesting membership`,

   // link rows
-  `<p><strong>${link.label}:</strong> <a href="${link.url}">${link.url}</a></p>`
+  `<p><strong>${link.label}:</strong> <a href="${escapeHtml(link.url)}">${escapeHtml(link.url)}</a></p>`

   // body
-        <p><strong>Name:</strong> ${displayName}</p>
-        <p><strong>Email:</strong> ${userEmail}</p>
+        <p><strong>Name:</strong> ${escapeHtml(displayName)}</p>
+        <p><strong>Email:</strong> ${escapeHtml(userEmail)}</p>
```

> **`escapeHtml` alone does NOT make the `href` safe.** A `javascript:`/`data:` URL contains no HTML-special characters, so it passes through escaping unchanged and executes when the admin clicks it. What neutralizes it is the URL-scheme validation below (Fix 2), which rejects any non-`http(s)` scheme *before* the email is built. **Fix 1 and Fix 2 must be committed together** — Fix 1 is not complete without Fix 2.

**Companion test is required** (per the project TDD standard — every `lib/utils/*` file has one):

```diff
// lib/utils/html.test.ts (new file)
+import { describe, it, expect } from 'vitest';
+import { escapeHtml } from './html';
+
+describe('escapeHtml', () => {
+  it('escapes all five HTML-significant characters', () => {
+    expect(escapeHtml(`<a href="x" onclick='y'>&`)).toBe(
+      '&lt;a href=&quot;x&quot; onclick=&#x27;y&#x27;&gt;&amp;',
+    );
+  });
+  it('escapes & first to avoid double-encoding', () => {
+    expect(escapeHtml('&lt;')).toBe('&amp;lt;');
+  });
+  it('leaves safe strings untouched', () => {
+    expect(escapeHtml('https://github.com/user')).toBe('https://github.com/user');
+  });
+});
```

Alternatives: scheme-check + inline-escape only (narrower), or switch the email to plain-text (`text:` instead of `html:`, drops the styled CTA).

**Follow-up (verified, separate PR):** `app/api/bug-report/route.ts` on `main` has the identical raw-HTML-email flaw — `${reporter}`, `${description}`, `${steps}`, `${url}` are interpolated raw into `html:`, and `${reporter}` into `subject:`/`replyTo:`. The same `escapeHtml` helper fixes it. Not in this PR's scope, but should be tracked as its own security fix so the new util gets reused, not re-implemented.
</details>

### ⚠️ Should-fix — missing URL validation (covers 2 findings)

`app/onboarding/actions.ts` only checks that a URL field is non-empty after `trim()`. It never validates the scheme or format, so:
- a `javascript:` URI passes, is stored, and renders as an `href` in the admin approvals UI (**security**), and
- a value like `not-a-url` passes and renders as a broken link (**dual-layer-validation convention violation**).

The server action is directly callable, so browser `type="url"` validation is not sufficient on its own.

<details><summary>Fix — server-side scheme validation (resolves both)</summary>

```diff
+function validateHttpUrl(url: string | null): string | null {
+  if (!url) return null;
+  try {
+    const parsed = new URL(url);
+    if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+      return `Please provide a valid URL starting with http:// or https://`;
+    }
+  } catch {
+    return `Please provide a valid URL (e.g. https://github.com/username)`;
+  }
+  return null;
+}

+  const urlValidationErrors = [
+    validateHttpUrl(linkedinUrl),
+    validateHttpUrl(githubUrl),
+    validateHttpUrl(portfolioUrl),
+  ].filter(Boolean);
+  if (urlValidationErrors.length > 0) {
+    return { error: urlValidationErrors[0] };
+  }
```

The `.filter(Boolean)` above is correct — `validateHttpUrl` returns an error *string* (truthy) or `null` (falsy), so nulls drop and errors surface. **This validation is load-bearing for the email `href` safety in the blocker fix** (it's what rejects `javascript:`/`data:` URLs), and it runs before the admin email is built — so ship it in the same commit as the escaping fix, not separately.
</details>

### ℹ️ Note — test mock nit

`app/onboarding/onboarding-form.test.tsx` wraps the mock variadically instead of assigning it directly, losing the typed signature.

```diff
 vi.mock('./actions', () => ({
-  completeOnboarding: (...args: unknown[]) => completeOnboardingMock(...args),
+  completeOnboarding: completeOnboardingMock,
 }));
```

---
*Generated by `/review-pr-fast` on 2026-07-02. 5 findings validated, 1 false positive filtered. AI_POLICY.md: this automated review assists but does not substitute human review.*

tonyrosario (changes requested):
See [comment](https://github.com/What-We-Will/community-platform/pull/187#issuecomment-4877793061) for requested changes

GeorgeDover:
@tonyrosario issues noted! Would you prefer if I just attended to the html injection issues related to this feature, or include `escapeHtml` coverage elsewhere too? Personally I think that would be better to tackle in a separate PR or when the areas are touched naturally, but up to you!

GeorgeDover:
@tonyrosario issues addressed now - I did have to change up the `const urlValidationErrors` filter logic to match the `OnboardingResult.error` type and also had to do some Vitest hoisting to avoid a TDZ `ReferenceError`, but otherwise kept the recommended changes 🙂

tonyrosario:
> @tonyrosario issues addressed now - I did have to change up the `const urlValidationErrors` filter logic to match the `OnboardingResult.error` type and also had to do some Vitest hoisting to avoid a TDZ `ReferenceError`, but otherwise kept the recommended changes 🙂

We can just address the blocker.

Suggest new utility:
```
/**
 * Escape the five HTML-significant characters to prevent markup/script injection.
 *
 * Output encoding for HTML **text** and **quoted** attribute values (for
 * example, a server-rendered nodemailer email body) — not HTML sanitization.
 * Escapes `&` `<` `>` `"` `'` (single quote as `&#39;`), replacing `&` first so
 * existing entities are not double-encoded.
 *
 * Not safe — do not use the output — in these contexts:
 * - Unquoted attributes (space, `=`, backtick unescaped) — keep values quoted.
 * - Inside `<script>`/`<style>` — entities aren't parsed; use `JSON.stringify`.
 * - URL in `href`/`src` — a `javascript:` scheme survives; validate separately.
 *
 * Not idempotent, so escape each raw value once. Non-string input (`null`,
 * `undefined`, numbers, objects) returns an empty string — it is never coerced,
 * so a mistyped value fails closed instead of leaking `"null"` into the output.
 *
 * @param value - The raw value to escape; coerced to a string.
 * @returns The escaped string, safe for HTML text or quoted-attribute contexts.
 * @example
 * // Neutralizes an injection payload:
 * escapeHtml('<img src=x onerror="alert(1)">');
 * // → "&lt;img src=x onerror=&quot;alert(1)&quot;&gt;"
 * @example
 * // Ordinary text with mixed quotes and ampersands:
 * escapeHtml(`Tom & "Jerry's" cat`);
 * // → "Tom &amp; &quot;Jerry&#39;s&quot; cat"
 */
export function escapeHtml(value: unknown): string {
  if (typeof value !== "string") return ""; // fail closed on non-string input
  return value
    .replace(/&/g, "&amp;") // must run first
    .replace(/</g, "&lt;")
    .replace(/>/g, "&gt;")
    .replace(/"/g, "&quot;")
    .replace(/'/g, "&#39;");
}
```

Test
```
import { escapeHtml } from "./html";

// The expected outputs below are the de-facto standard shared by escape-html,
// html-escaper, and lodash.escape (single quote → &#39;). Asserting the exact
// strings pins this helper to the reference behavior of those libraries.

describe("escapeHtml", () => {
  describe("the five HTML-significant characters", () => {
    it("escapes &", () => {
      expect(escapeHtml("&")).toBe("&amp;");
    });
    it("escapes <", () => {
      expect(escapeHtml("<")).toBe("&lt;");
    });
    it("escapes >", () => {
      expect(escapeHtml(">")).toBe("&gt;");
    });
    it("escapes the double quote (double-quoted attribute context)", () => {
      expect(escapeHtml('"')).toBe("&quot;");
    });
    it("escapes the single quote as &#39; not &apos; (single-quoted attribute context)", () => {
      expect(escapeHtml("'")).toBe("&#39;");
    });
  });

  describe("ordering — & is escaped first, no double-encoding", () => {
    it("escapes a literal ampersand in entity-looking text", () => {
      // A single pass: only the & is HTML-significant here.
      expect(escapeHtml("&lt;")).toBe("&amp;lt;");
    });
    it("escapes a run of ampersands without cascading", () => {
      expect(escapeHtml("a & b & c")).toBe("a &amp; b &amp; c");
    });
    it("escapes an ampersand adjacent to a real angle bracket", () => {
      expect(escapeHtml("Tom & <Jerry>")).toBe("Tom &amp; &lt;Jerry&gt;");
    });
    it("is NOT idempotent — re-escaping double-encodes (escape each raw value once)", () => {
      // Pins the double-encoding so a contributor who wraps an already-escaped
      // string gets a failing test rather than silently garbled email copy.
      expect(escapeHtml("a &amp; b")).toBe("a &amp;amp; b");
    });
  });

  describe("real-world injection payloads are neutralized", () => {
    it("script tag", () => {
      expect(escapeHtml("<script>alert(1)</script>")).toBe(
        "&lt;script&gt;alert(1)&lt;/script&gt;",
      );
    });
    it("img/onerror with double-quoted handler", () => {
      expect(escapeHtml('<img src=x onerror="alert(1)">')).toBe(
        "&lt;img src=x onerror=&quot;alert(1)&quot;&gt;",
      );
    });
    it("attribute breakout via quotes (double and single)", () => {
      expect(escapeHtml(`" onmouseover='alert(1)'`)).toBe(
        "&quot; onmouseover=&#39;alert(1)&#39;",
      );
    });
    it("neutralizes the HTML comment terminator (regression lock on > escaping)", () => {
      // The `>` escape is the only thing preventing comment-context injection;
      // lock it so a future contributor can't drop it unnoticed.
      expect(escapeHtml("-->")).toBe("--&gt;");
      expect(escapeHtml("--!>")).toBe("--!&gt;");
    });
  });

  describe("passthrough and non-string handling", () => {
    it("leaves a safe string untouched", () => {
      expect(escapeHtml("https://github.com/octocat")).toBe(
        "https://github.com/octocat",
      );
    });
    it("passes backtick through unchanged — deliberate; UNSAFE in unquoted attributes", () => {
      // Backtick is not HTML-significant in standards-compliant renderers. Pinned
      // so nobody adds a backtick escape without revisiting the unquoted-attr rule.
      expect(escapeHtml("`")).toBe("`");
    });
    it("returns empty string for empty input", () => {
      expect(escapeHtml("")).toBe("");
    });
    it("returns empty string for null and undefined — fails closed, does not coerce", () => {
      expect(escapeHtml(null)).toBe("");
      expect(escapeHtml(undefined)).toBe("");
    });
    it("returns empty string for numbers, objects, and arrays (never String()-coerced)", () => {
      expect(escapeHtml(42)).toBe("");
      expect(escapeHtml({})).toBe("");
      expect(escapeHtml(["<b>"])).toBe("");
    });
  });
});
```

### 3 Gaps to fix

| Priority | File | Values to escape | Threat |
|----------|------|------------------|--------|
| 🔴 **Highest** | `app/api/messages/route.ts` | `senderName` (subject **and** body), `previewText` (message body) | **user → user** — a member DMs another member `` and it renders in the recipient's email client |
| 🟠 | `app/onboarding/actions.ts` | `displayName` (subject **and** body), `userEmail` (body) | user → admin (the original PR-187 finding) |
| 🟠 | `app/api/bug-report/route.ts` | `reporter`, `description`, `steps`, `url` | user → admin (also tracked in M16) |

tonyrosario:
> @tonyrosario issues addressed now - I did have to change up the `const urlValidationErrors` filter logic to match the `OnboardingResult.error` type and also had to do some Vitest hoisting to avoid a TDZ `ReferenceError`, but otherwise kept the recommended changes 🙂

Looks like we were working in parallel... will take another look 👀

app/onboarding/actions.ts
```diff
@@ -4,9 +4,23 @@ import { revalidatePath } from "next/cache";
 import { createClient } from "@/lib/supabase/server";
 import nodemailer from "nodemailer";
 import { safeTimezone } from "@/lib/utils/timezone";
+import { escapeHtml } from "@/lib/utils/html";
 
 export type OnboardingResult = { error?: string };
 
+function validateHttpUrl(url: string | null): string | null {
+  if (!url) return null;
```
tonyrosario: Few things here that I noticed:

1. The method is named `validate` but can return null.  I would expect a boolean as a return type if we're validating something; aka it is or isn't valid.

2. Method has 4 exits which seems strange for such a small function

3. We should only allow "https" (drop "http")

4. the call site returns only the first error message (currently line 66) even if there are multiple.
  - Question: would we ever want to identify which of the 3 urls are invalid?  (linkedin, github, portfolio and all permutations)

* * * 
The core intent is right but the observations above are signs of low cohesion, one function doing multiple things (presence, validation, error messages).

Suggest:
1. Extracting the validation to a method.

New utility for url (I know we just added one for html! 😭) but it needs its own 'home' (SRP) and I suspect that we will be needing more utils around urls in the future.

`lib/utils/url.ts`
```
  function isHttpsUrl(url: string): boolean {
    try {
      return new URL(url).protocol === "https:";
    } catch {
      return false;
    }
  }
```
2. Let the calling code own presence + messaging.

After the validation is extracted, the other two concerns can move up to the caller:
  - Presence: an empty optional field isn't invalid, just absent, so it can be skipped
  - Which field(s) is/are wrong: goes back to #4 above.

Up to you how you want to

app/onboarding/onboarding-form.test.tsx
```diff
@@ -9,10 +10,12 @@ global.ResizeObserver = class {
...
 };
 
-const completeOnboardingMock = vi.fn();
+const completeOnboardingMock = vi.hoisted(() =>
+  vi.fn<typeof completeOnboarding>()
+);
 
 vi.mock("./actions", () => ({
-  completeOnboarding: (...args: unknown[]) => completeOnboardingMock(...args),
+  completeOnboarding: completeOnboardingMock,
```
tonyrosario, line 22: LGTM 👍

app/onboarding/actions.ts
GeorgeDover: Understood. I was following the validate function [you prescribed earlier](https://github.com/What-We-Will/community-platform/pull/187#issuecomment-4877793061), but can address those concerns with it. I think the [testing gaps and lack of function tooltip](https://github.com/What-We-Will/community-platform/pull/187#issuecomment-4879177231) are very valid points and will incorporate those for sure. And yes we definitely should enforce `https` - I just wasn't sure if we were already doing that or not!

kaizengrowth (commented):
On a functional test in local, I can verify that this works as expected. I can create a user without a LinkedIn link.

xdhmoore:
I have two comments:

1. I strongly suggest we use an existing escaping library instead of trying to roll our own, especially since this is a security issue. I see a couple libraries that might work: [DOMPurify](https://github.com/cure53/DOMPurify) or [sanitize-html](https://www.npmjs.com/package/sanitize-html). It looks like DOMPurify may have more recent development. Generally speaking, escaping that needs to be performed can be different depending on the context. For example, variables rendering in the middle of text vs in an HTML attribute like src vs in CSS. The escaping needing for these contexts is different and a single function broadly applied will not cover all cases. So, we need a library to help with that. The library should give us the option to specify the context so it can apply the correct escaping. 

Though, I'm a little surprised React isn't escaping server-side. I believe it normally does some escaping. That may be worth looking into if we haven't already.

2. But even aside from the code injection issue, do we really want arbitrary users sending arbitrary links to an admin? For example, what if the domain is a personal website, and the admin clicks it only to be taken to a malware site?

I think we should limit the domains that a user can enter. Sure, LinkedIn alone may not be enough, but let's limit it to a set number of domains that we know are safe: stackoverflow, github, facebook, gitlab, bitbucket, etc. Personal websites, while nice, are too much of a security risk in my opinion.

GeorgeDover:
> I have two comments:
> 
> 1. I strongly suggest we use an existing escaping library instead of trying to roll our own, especially since this is a security issue. I see a couple libraries that might work: [DOMPurify](https://github.com/cure53/DOMPurify) or [sanitize-html](https://www.npmjs.com/package/sanitize-html). It looks like DOMPurify may have more recent development. Generally speaking, escaping that needs to be performed can be different depending on the context. For example, variables rendering in the middle of text vs in an HTML attribute like src vs in CSS. The escaping needing for these contexts is different and a single function broadly applied will not cover all cases. So, we need a library to help with that. The library should give us the option to specify the context so it can apply the correct escaping.
> 
> Though, I'm a little surprised React isn't escaping server-side. I believe it normally does some escaping. That may be worth looking into if we haven't already.
> 
> 2. But even aside from the code injection issue, do we really want arbitrary users sending arbitrary links to an admin? For example, what if the domain is a personal website, and the admin clicks it only to be taken to a malware site?
> 
> I think we should limit the domains that a user can enter. Sure, LinkedIn alone may not be enough, but let's limit it to a set number of domains that we know are safe: stackoverflow, github, facebook, gitlab, bitbucket, etc. Personal websites, while nice, are too much of a security risk in my opinion.

I think this a wise tactic. I don't have much free time this week and unfortunately didn't have time to get around to it last weekend, so anyone can feel free to implement this plan! DOMPurify is something I used before and liked it, but I didn't want to deviate from our previous pattern without consensus. But it looks like the approach I went with here wasn't bold enough so I think it's justified now. @xdhmoore @kaizengrowth @tonyrosario

kaizengrowth:
I'll make an issue ticket for Daniel's comment, so we can pursue this in a separate PR.

app/admin/approvals/page.tsx
```diff
@@ -39,7 +39,7 @@ export default async function ApprovalsPage() {
         <div>
           <h1 className="text-2xl font-semibold">Pending Approvals</h1>
           <p className="mt-1 text-sm text-muted-foreground">
-            Review LinkedIn profiles and approve tech workers to join the platform.
+            Review applicants&apos; LinkedIn, GitHub, or website links and approve tech workers to join the platform.
```
xdhmoore, line 43: Nit:
Suggestion for readability:
> Review LinkedIn, GitHub, or other applicant-provided links...

app/onboarding/actions.ts
```diff
@@ -91,15 +131,28 @@ async function notifyAdminOfNewApplication({
...
+    const links = [
+      { label: "LinkedIn", url: linkedinUrl },
+      { label: "GitHub", url: githubUrl },
+      { label: "Website", url: portfolioUrl },
+    ].filter((link): link is { label: string; url: string } => Boolean(link.url));
+
+    const linksHtml = links
+      .map(
+        (link) =>
+          `<p><strong>${link.label}:</strong> <a href="${escapeHtml(link.url)}">${escapeHtml(link.url)}</a></p>`
```
xdhmoore, line 141: So, please correct me if I'm wrong. I'm new to the codebase but it looks to me like the real issue here is that we are not using React at these points to render the HTML. If we put this in a tsx page and use JSX instead of js template strings, React will do this escaping for us:
https://reactz2h.com/chapter_12_frontend_security_for_react/series_01_xss_and_injection_defense/understanding_react_escaping 

I realize that we are rendering an email though server-side and that may not be setup. Is rendering these emails with react doable?

xdhmoore:
@tonyrosario is that automated security review something that's an automated part of the review pipeline, or are you running it manually? Just wondering if someone else is able to run it again after fixing the code.

tonyrosario:
I made an ADR to document the decision of bringing in those libraries (DOMPurify or sanitize-html) and the proposal stands as 'rejected' **for this specific use-case**.  I am open to changing it but I don't think it's worth the time/tokens.

Those tools are used for user-authored HTML but this feature accepts URLs as strings, no markup.

Having said that, I tried to purposefully tweak the wording of the ADR so that it does NOT lock-out bringing in these kinds of tools in the future; this should only cover plain URL-string validation.
(btw ChatGPT did better than Claude for editing the ADR)

@xdhmoore @GeorgeDover @kaizengrowth 

EDIT: holding merge until @xdhmoore can take a look 👀

tonyrosario:
> @tonyrosario is that automated security review something that's an automated part of the review pipeline, or are you running it manually? Just wondering if someone else is able to run it again after fixing the code.

sorry - it seems I missed this.

yes, this is something I can add to the project.  I have 2 versions, one is more token conservative and the other is not (faster / parallel).  Working on cleaning it up before adding it to the project or another internal app.

app/onboarding/actions.ts
tonyrosario: We're all learning here 🧑‍🏫  so I appreciate your perspective!

I'm just trying to get George's PR merged at this point ✅ 

I had an agent test our code, React (19.2.3), and go these results... which your read is correct: React would take over text and quoted-attribute escaping here, and `escapeHtml` could go away.

| Input | Output |
|---|---|
| text `` | `&lt;img src=x onerror=alert(1)&gt;` |
| `href='https://x.com/"><script>'` | `href="https://x.com/&quot;&gt;&lt;script&gt;"` |
| `href="javascript:alert(1)"` | `href="javascript:throw new Error('React has blocked a javascript: URL…')"` |
| `href="data:text/html,<script>alert(1)</script>"` | **unchanged** |

However, the two things it doesn't cover are the two controls this PR actually added:

1. **No URL scheme validation.** `data:` passes straight through, and
   `javascript:` isn't rejected — it's rewritten into a different, inert
   `javascript:` URL. That's a browser-DOM safety net, not a validator, and it
   isn't a substitute for `isHttpsUrl`.
2. **`subject` isn't HTML.** It's an SMTP header, so React has no reach into
   it. That one needs CRLF stripping, and entity-encoding it is actively wrong
   (entities render literally in the inbox) — which is what the second commit
   in this PR fixes.

By zooming out, the review found four separate sendMail call sites (onboarding, approvals, messages, bug reports), and **each one hand-rolls** its own transport config and HTML! Your React suggestion would be a good answer to it but I'd ask that we create a separate issue for it and not include it in this PR.

_(Input/Output table + issues 1 + 2 were AI generated)_

